PR #452, opened 2026-04-07
fix: reality checkers no longer persist state to disk during refresh

### LINKED ISSUE #424: Reality checkers doesn't need to update state during refresh
Currently reality checker persists state every time it checks reality. We need to ensure state persistence is called explicitly by RSL or upper layer without Reality checkers updating state on disk.

leninmehedy:
## Description

Reality checkers no longer persist state to disk. State persistence is now the responsibility of RSL/upper layers.

### Related Issues

* Closes #424


brunodam (dismissed):
LGTM

brunodam (approved):
LGTM

PR #488, opened 2026-05-09
feat(security): auto-create system accounts and fix ownership model

### LINKED ISSUE #394: /opt/solo/weaver directory should be owned by user id 2500
Currently, every file and folder created by weaver under /opt/solo/weaver is owned by root user.
It should be owned by the user id 2500 which is the user created for weaver purposes.

Image

**Actually, let's change the default user to hedera, gid and uid 2000 instead 2500. We can add an option to allow overwriting that user.**

### LINKED ISSUE #404: Remove panic errors from security package
We need ensure there is no method that panics: https://github.com/hashgraph/solo-weaver/blob/main/pkg/security/principal/manager.go#L46-L64

brunodam:
## Description

This pull request implements a new two-user security model for the Solo Provisioner, improves directory and file ownership handling, and introduces a new system daemon mode. The changes ensure that block-node storage directories are correctly owned and accessible by both the provisioner daemon and block-node containers, and that the daemon runs as a non-root user with appropriate permissions. It also adds a new `daemon` command and improves logging setup and reporting.

**Security Model and Directory Ownership Improvements:**

* Implements a two-user security model: the provisioner now runs as `weaver:2500`, and block-node storage directories are owned by `hedera:2000`, with setgid directories to ensure correct group inheritance and access. This is documented in a new design doc and implemented across several files, including new hardcoded storage owner identity, directory permission changes, and group membership management.
* Removes runtime creation of the `weaver` user and group from CI workflows, as user setup is now handled at install time, not during test or boot. (`.github/workflows/zxc-unit-test.yaml`, `.github/workflows/zxc-integration-test.yaml`, `.github/workflows/zxc-uat-test.yaml`) [[1]](diffhunk://#diff-16519657be7a7de47f7810334df1f22e9ab4ab0fe4f3f3bbe13bb61e73eea922L72-L76) [[2]](diffhunk://#diff-68122e070f3aff4b7ec9fcb448483b386d6e4218697f020183c14e30257df2cdL182-L183) [[3]](diffhunk://#diff-8deb24091ab41d19518c4d0473ecc4780c16568e6924166c28ac367d3bfa69feL174-L175)

**Daemon Mode and Logging Enhancements:**

* Adds a new `daemon` command to run the Solo Provisioner as a long-lived system daemon, with proper signal handling and integration into the root command. (`cmd/weaver/commands/daemon.go`, `cmd/weaver/commands/root.go`) [[1]](diffhunk://#diff-cfb63ebe4e4926fab208f0ee9eca46821bd2dc2c891d819badeea316672919ceR1-R40) [[2]](diffhunk://#diff-45410074f2fd3857e628ffaa910cde1e59b5323af44d8ddb2516799b4916d62bR97)
* Updates logging configuration to set log file and directory ownership and permissions correctly, including pre-creating log files with the right group and mode, and using a different log filename for the daemon process. (`cmd/weaver/commands/root.go`)

**User Experience and Reporting:**

* Updates the workflow report summary to include the daemon log path for better traceability in logs and reports. (`cmd/weaver/commands/common/run.go`)
* Adds guidance to the `CLAUDE.md` contributor doc to require explicit user approval before committing or pushing changes. (`CLAUDE.md`)

### Related Issues

* Closes #394
* Closes #404


internal/bll/blocknode/install_handler.go
```diff
@@ -63,10 +63,18 @@ func (h *InstallHandler) BuildWorkflow(
 	var wb *automa.WorkflowBuilder
 	if currentState.ClusterState.Created {
 		wb = automa.NewWorkflowBuilder().WithId("block-node-install").
-			Steps(steps.SetupBlockNode(ins))
+			Steps(
+				steps.EnsureWeaverOwnerStep(),
```
brunodam, line 72: Shouldn't this be EnsureHederaUserStep()? Perhaps another workflow?

internal/templates/files/weaver/solo-provisioner.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Solo Provisioner Daemon
+After=network.target
+
+[Service]
+Type=simple
+User=weaver
+Group=weaver
+ExecStart=/opt/solo/weaver/bin/solo-provisioner daemon
```
brunodam, line 9: This daemon does not exist yet. So, systemctl status will probably show errors. Let's avoid that and disable the step that installs this for now.

leninmehedy (dismissed):

internal/bll/blocknode/install_handler.go
brunodam: This is actually a backwards-compatibility guard rather than block-node setup logic.

The hedera user is already handled — EnsureHederaOwnerStep runs as the first step inside SetupBlockNode, which is the right place for it since hedera ownership is purely a block-node storage concern.

Older released binaries did not include EnsureWeaverOwnerStep in their self-install workflow (NewSelfInstallWorkflow). This means machines installed with an old binary may have the provisioner binary on disk but no weaver:2500 account. The global pre-run check (RunPersistentPreRun) only validates that the binary exists, not that the user account exists. So without this step, a user who installed with an old binary and then runs block node install with a new binary would hit a failure in InstallClusterWorkflow's preflight check that validates weaver:2500.

In other words: it's not here because block node needs the weaver user — it's here because block node install may be the first command run with a new binary on a machine that was provisioned before weaver's self-install created the account. The step is idempotent so it's a no-op on up-to-date machines.

internal/workflows/weaver.go
```diff
@@ -17,14 +17,19 @@ func CheckWeaverInstallationWorkflow() *automa.WorkflowBuilder {
 func NewSelfInstallWorkflow() *automa.WorkflowBuilder {
 	return automa.NewWorkflowBuilder().WithId("self-install-workflow").Steps(
 		CheckPrivilegesStep(),
+		steps.EnsureWeaverOwnerStep(),
 		steps.SetupHomeDirectoryStructure(models.Paths()),
 		steps.InstallWeaver(models.Paths().BinDir),
+		steps.InstallSudoersStep(),
+		steps.InstallWeaverServiceStep(),
```
leninmehedy: Are we sure we want to install the daemon? Since the daemon doesn't do anything... if we install now and later modify, we shall need to handle the migration.

leninmehedy (dismissed):

leninmehedy (approved):

Issue #520, opened 2026-05-14
Story: Implement soak-watcher goroutine for migration soak criteria monitoring and auto-decommission


leninmehedy:
Part of epic #499 — solo-provisioner-daemon Core.

Implement soak-watcher goroutine: activated via `POST /soak/start`; monitors all migration soak criteria; auto-triggers decommission when all criteria are green; goroutine exits after decommission while daemon continues running.

Issue #519, opened 2026-05-14
Story: Implement upgrade-monitor goroutine watching NetworkUpgradeExecute CR for ReadyForProvisionerDaemon


leninmehedy:
Part of epic #499 — solo-provisioner-daemon Core.

Implement upgrade-monitor goroutine: continuously watches Kubernetes API for `NetworkUpgradeExecute` CR status `ReadyForProvisionerDaemon`; triggers the execute-phase workflow when detected.

PR #599, opened 2026-05-21
feat(daemon): implement Unix socket server with health and soak endpoints (#518)

### LINKED ISSUE #518: Story: Implement Unix socket server with /health, /soak/status, /soak/start, /upgrade/abort endpoints
Part of epic #499 — solo-provisioner-daemon Core.

Implement Unix socket server at `/opt/solo/weaver/provisioner/daemon.sock` with endpoints: `GET /health`, `GET /soak/status`, `POST /soak/start`, `POST /upgrade/abort`.

leninmehedy:
## Summary

- Introduces `internal/daemon` package structured as a **controller + sub-systems**: `Daemon` is a thin orchestrator composing `consensus.UpgradeMonitor`, `consensus.MigrationMonitor`, and `Server` so stories #519 and #520 can be implemented in isolation
- Implements Unix domain socket HTTP server with three endpoints: `GET /health`, `GET /migration/consensus/soak/status`, `POST /migration/consensus/soak/start`
- Extends `WeaverPaths` with `DaemonDir` and `DaemonSockPath` (derived from `DaemonDir`); `DaemonDir` appended to `AllDirectories` so it is auto-created on install
- Wires `cmd/daemon/main.go` to `daemon.New(models.Paths()).Run(ctx)`

Closes #518

## Architecture

```
internal/daemon/
  daemon.go          — thin controller; composes sub-systems via errgroup
  server.go          — Unix socket HTTP server (MkdirAll + Listen + Chmod)
  handlers.go        — route handlers
  types.go           — shared types: HealthResponse, ErrorResponse
  server_test.go
  consensus/         — consensus-node concerns; future components get siblings
    upgrade_monitor.go    — K8s watch stub (story #519)
    migration_monitor.go  — soak lifecycle: channel, status, goroutine (story #520)
    types.go              — SoakStartRequest/Response/Status + Validate()
    types_test.go
```

URL namespace is hierarchical — `/migration/consensus/soak/...` leaves room for `/upgrade/consensus/...`, `/migration/blocknode/...` etc. in future stories.

## Design highlights

- `Daemon.Run` is a 3-line `errgroup` orchestrator — server startup failure cancels the daemon context and surfaces the error
- `MigrationMonitor.TryEnqueue` uses `soakActive.Swap(true)` before channel send to close the race window between goroutine spawn and first `soakStatus` store
- `MigrationMonitor.Status()` returns `*SoakStatusResponse` (zero-copy; idle sentinel is a package-level var)
- `MigrationMonitor.run` uses a single outermost `defer` — recovery wraps all cleanup to prevent LIFO panic shadowing
- `defer soakWg.Wait()` inside `MigrationMonitor.Run` — fully quiesced daemon on return
- `Server.Start` calls `os.MkdirAll` on the socket parent before `net.Listen` — first run on a fresh install never fails with "no such file or directory"
- Socket permissions set to `0o660` after `net.Listen` — prevents unprivileged local access
- `http.Server.Shutdown` uses a 5 s timeout context — cannot hang indefinitely
- `http.MaxBytesReader(16 KiB)` caps request bodies in `handleSoakStart`
- `ServerConfig.ReadHeaderTimeout` is injectable — test uses 50 ms, production defaults to 5 s
- All state-transition log entries include a `reason` field per HIP observability schema
- Migration plan path convention: `/opt/solo/weaver/migration/consensus/<node-id>-<ISO8601>-migration-plan.yaml`

## Test plan

- [ ] `task lint` — passes locally
- [ ] `task vm:test:unit` — run in UTM VM
- [ ] Manual smoke (verified on UTM VM):
  ```
  sudo curl --unix-socket /opt/solo/weaver/daemon/daemon.sock http://local/health
  # → {"status":"ok"}
  sudo curl --unix-socket /opt/solo/weaver/daemon/daemon.sock http://local/migration/consensus/soak/status
  # → {"active":false}
  curl --unix-socket /opt/solo/weaver/daemon/daemon.sock http://local/health
  # → connection refused (0o660 permission check working)
  ```

Review guide: `docs/claude/reviews/00518-implement-unix-socket-server-with-health.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/claude/reviews/00518-implement-unix-socket-server-with-health.md
+++ b/docs/claude/reviews/00518-implement-unix-socket-server-with-health.md
@@ -74,8 +74,7 @@ task lint
 
 ```bash
 # 1. Build CLI and daemon binaries
-task build:cli GOOS=linux GOARCH=arm64
-task build:weaver GOOS=linux GOARCH=arm64
+task build:daemon GOOS=linux GOARCH=arm64
 
 # 2. Install solo-provisioner to create the directory structure
 sudo ./solo-provisioner-linux-arm64 install
--- a/docs/claude/reviews/00518-implement-unix-socket-server-with-health.md
+++ b/docs/claude/reviews/00518-implement-unix-socket-server-with-health.md
@@ -1,0 +1,113 @@
+# Review guide — #518 Unix socket server
+
+> **Branch:** `00518-implement-unix-socket-server-with-health`
+> **PR base:** `00499-feat-solo-provisioner-daemon-core`
+> **Issue:** https://github.com/hashgraph/solo-weaver/issues/518
+
+## Problem and solution
+
+No `internal/daemon/` package existed. This PR adds the Unix domain socket HTTP server
+that forms the control plane for `solo-provisioner-daemon`. The daemon is structured as a
+thin controller (`Daemon`) composing three sub-systems (`Server`, `UpgradeMonitor`, `SoakWatcher`)
+so stories #519 and #520 can implement their respective types in isolation.
+
+## Changed files
+
+| File | Description |
+|---|---|
+| `pkg/models/weaver_paths.go` | Add `DaemonDir`, `DaemonSockPath` (derived from `DaemonDir`); append to `AllDirectories` |
+| `internal/daemon/types.go` | `HealthResponse`, `ErrorResponse`, `SoakStartRequest`, `SoakStartResponse`, `SoakStatusResponse` |
+| `internal/daemon/upgrade_monitor.go` | New: `UpgradeMonitor` stub; story #519 implements `Run(ctx)` here |
+| `internal/daemon/soak_watcher.go` | New: `SoakWatcher` — all soak state, `TryEnqueue`, `Status`, `Run`, `run`, `resumeIfNeeded` |
+| `internal/daemon/daemon.go` | New: thin controller; `Run` is a 3-line errgroup orchestrator |
+| `internal/daemon/server.go` | New: Unix socket server; holds `sockPath string` + `*SoakWatcher` |
+| `internal/daemon/handlers.go` | New: `handleHealth`, `handleSoakStatus`, `handleSoakStart` |
+| `internal/daemon/server_test.go` | New: 6 unit tests over a real unix socket in a temp dir |
+| `cmd/daemon/main.go` | Wire existing entrypoint to `daemon.New(models.Paths()).Run(ctx)` |
+| `docs/claude/plans/00518-…` | Plan doc (architecture decisions, observability scope) |
+
+## Code review checklist
+
+- [ ] `DaemonSockPath` is derived from `pp.DaemonDir` in post-init block (not a repeated string literal)
+- [ ] `DaemonDir` appears in `AllDirectories` — confirms auto-creation on install
+- [ ] `Daemon.Run` is a pure orchestrator — no state, no channels directly on `Daemon`
+- [ ] `Server` holds `sockPath string` + `*SoakWatcher` + `*http.Server` — not coupled to `Daemon`
+- [ ] `http.Server` has `ReadHeaderTimeout: 5s` set
+- [ ] `errors.Is(err, http.ErrServerClosed)` used (not `!=`)
+- [ ] `SoakWatcher.TryEnqueue` uses `soakActive.Swap(true)` before channel send — closes the race window
+- [ ] `SoakWatcher.Status()` returns `*SoakStatusResponse` (pointer, not copy); idle sentinel is package-level var
+- [ ] `SoakWatcher.run` has a single outermost `defer` — recovery wraps all cleanup (no LIFO panic shadowing)
+- [ ] `soakWg.Wait()` is deferred inside `SoakWatcher.Run` — quiescence guaranteed before return
+- [ ] `soakWg.Add(1)` called before goroutine spawn in dispatch loop
+- [ ] `resumeIfNeeded` stub documents `soakActive.Store(true)` + `soakWg.Add(1)` invariant for story #520
+- [ ] `os.Chmod(0o660)` called after `net.Listen`; chmod failure closes listener, removes socket, and aborts
+- [ ] `Shutdown` uses `context.WithTimeout(5s)` — cannot hang indefinitely
+- [ ] `errgroup.WithContext` used — server startup failure cancels daemon context
+- [ ] `handleSoakStart` caps body with `http.MaxBytesReader(16 KiB)` and validates `node_id` is non-empty
+- [ ] All error responses use `writeError` → `ErrorResponse{Error string}` with `Content-Type: application/json`
+- [ ] `SoakStartResponse{Accepted bool}` used (not `map[string]bool`)
+- [ ] All state-transition log entries include `reason` field
+- [ ] No `/upgrade/abort` route — correct per design
+- [ ] `server_test.go` uses `require.Eventually` for async soak status polling — no flaky sleep
+- [ ] Conflict test asserts `Content-Type: application/json` and decodes `ErrorResponse`
+- [ ] `Test_SoakStart_InvalidBody` and `Test_SoakStart_MissingNodeID` cover 400 paths
+- [ ] Build tag `!integration` on test file — safe for `task test:unit`
+
+## Test commands
+
+```bash
+# Run daemon package unit tests (in UTM VM)
+task vm:test:unit
+
+# Targeted coverage
+task test:coverage TEST_PATHS=./internal/daemon/... TEST_REGEX="."
+
+# Lint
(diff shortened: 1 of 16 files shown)


internal/daemon/server.go
```diff
@@ -0,0 +1,78 @@
...
+	select {
+	case <-ctx.Done():
+		logx.As().Info().Str("reason", "ServerStopped").Msg("Daemon socket server shutting down")
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		shutdownErr := s.srv.Shutdown(shutdownCtx)
+		_ = os.Remove(sockPath)
+		return shutdownErr
+	case err := <-serveErr:
+		_ = os.Remove(sockPath)
```
brunodam, line 101:
```suggestion
		logx.As().Error().Err(err).Str("reason", "ServerStopped").Msg("Daemon socket server exited with error")
		_ = os.Remove(sockPath)
```

internal/daemon/server.go
```diff
@@ -0,0 +1,78 @@
...
+
+func NewServer(sockPath string, sw *SoakWatcher) *Server {
+	s := &Server{sockPath: sockPath, sw: sw}
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("GET /health", s.handleHealth)
+	mux.HandleFunc("GET /soak/status", s.handleSoakStatus)
+	mux.HandleFunc("POST /soak/start", s.handleSoakStart)
+
+	s.srv = &http.Server{Handler: mux}
```
brunodam: should we add a read header timeout here just in case?


```suggestion
	s.srv = &http.Server{Handler: mux, ReadHeaderTimeout: 5 * time.Second}
```

internal/daemon/handlers.go
```diff
@@ -0,0 +1,50 @@
...
+func (s *Server) handleSoakStatus(w http.ResponseWriter, _ *http.Request) {
+	writeJSON(w, http.StatusOK, s.sw.Status())
+}
+
+func (s *Server) handleSoakStart(w http.ResponseWriter, r *http.Request) {
+	var req SoakStartRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		writeError(w, http.StatusBadRequest, "invalid request body")
+		return
+	}
```
brunodam: should we have a body-size cap and minimal field validation here?

Something like:

```
  r.Body = http.MaxBytesReader(w, r.Body, 16<<10) // 16 KiB
  var req SoakStartRequest
  if err := json.NewDecoder(r.Body).Decode(&req); err != nil { ... }
  if req.NodeID == "" {
      writeError(w, http.StatusBadRequest, "node_id is required")
      return
  }
```

internal/daemon/server.go
```diff
@@ -0,0 +1,78 @@
...
+		_ = ln.Close()
+		_ = os.Remove(sockPath)
+		return err
+	}
+
+	logx.As().Info().Str("reason", "ServerStarted").Str("sock", sockPath).Msg("Daemon socket server listening")
+
+	serveErr := make(chan error, 1)
+	go func() {
+		if err := s.srv.Serve(ln); err != nil && err != http.ErrServerClosed {
```
brunodam: Should we use errors.Is for http.ErrServerClosed instead?


```suggestion
		if err := s.srv.Serve(ln); err != nil && !errors.Is(err, http.ErrServerClosed) {
```

internal/daemon/handlers.go
```diff
@@ -0,0 +1,50 @@
...
+
+	logx.As().Info().
+		Str("reason", "SoakStartAccepted").
+		Str("node_id", req.NodeID).
+		Str("migration_plan", req.MigrationPlanPath).
+		Time("cutover_ts", req.CutoverTimestamp).
+		Msg("Soak start request accepted")
+
+	writeJSON(w, http.StatusAccepted, map[string]bool{"accepted": true})
```
brunodam: I see why we used map[string]bool given how simple the reply is.

I wonder if it would be worth to have a type for it in the types.go file as well.

something like:

```
  // SoakStartResponse is returned by POST /soak/start on accept.
  type SoakStartResponse struct {
      Accepted bool `json:"accepted"`
  }
```

internal/daemon/server_test.go
```diff
@@ -0,0 +1,159 @@
...
+	require.Eventually(t, func() bool {
+		var status daemon.SoakStatusResponse
+		getJSON(t, client, "http://daemon/soak/status", &status)
+		return status.Active
+	}, 2*time.Second, 10*time.Millisecond)
+
+	// Second POST while watcher is running should be rejected.
+	resp2, err := client.Post("http://daemon/soak/start",
+		"application/json", bytes.NewReader(mustMarshal(t, payload)))
+	require.NoError(t, err)
```
brunodam, line 164: should we also add a body shape assertion:



```suggestion
	require.NoError(t, err)
	
	var errBody daemon.ErrorResponse
	require.NoError(t, json.NewDecoder(resp2.Body).Decode(&errBody))
	assert.NotEmpty(t, errBody.Error)
	assert.Equal(t, "application/json", resp2.Header.Get("Content-Type"))
```

internal/daemon/server_test.go
```diff
@@ -0,0 +1,159 @@
...
+	defer resp2.Body.Close()
+	assert.Equal(t, http.StatusConflict, resp2.StatusCode)
+}
+
```
brunodam, line 202: Shoud we add a test to cover the 400 path as well?


```suggestion
  func Test_SoakStart_InvalidBody(t *testing.T) {
      client, cancel := startTestDaemon(t)
      defer cancel()

      resp, err := client.Post("http://daemon/soak/start",
          "application/json", bytes.NewReader([]byte("not-json")))
      require.NoError(t, err)
      defer resp.Body.Close()
      assert.Equal(t, http.StatusBadRequest, resp.StatusCode)
  }
```

brunodam (commented):
LGTM. 

Just a few nitpicks that can be fixed later if you prefer.

brunodam (approved):

docs/claude/reviews/00518-implement-unix-socket-server-with-health.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+## Manual UAT (UTM VM)
+
+```bash
+# 1. Build the daemon binary
+task build:weaver GOOS=linux GOARCH=amd64
+
+# 2. Copy to VM and run
+sudo ./solo-provisioner-daemon &
```
brunodam: I've noticed that first we need to build and install solo-provisioner CLI so that the directory structure gets created.


```suggestion
# 1. Build the CLI binary
task build:cli GOOS=linux GOARCH=arm64

# 2. Install solo-provisioner so that the directory structure gets created first
sudo ./solo-provisioner-linux-arm64 install

# 3. Build the daemon binary
task build:weaver GOOS=linux GOARCH=arm64

# 2. Copy to VM and run
sudo ./solo-provisioner-daemon-linux-arm64 &
```

docs/claude/reviews/00518-implement-unix-socket-server-with-health.md
```diff
@@ -0,0 +1,98 @@
...
+
+# 3. Health check
+curl --unix-socket /opt/solo/weaver/daemon/daemon.sock http://local/health
```
brunodam: I've noticed I had to use sudo in order for this to work.

Am I right to think that it will be a hedera or root user that will try to see that daemon.sock file. If so, I believe we should be ok.

brunodam (approved):

docs/claude/reviews/00518-implement-unix-socket-server-with-health.md
leninmehedy: Yes, it should be weaver or root. Since `solo-provisioner` CLI will need that level access to operate.

leninmehedy: daemon runs as the weaver group service user (or root in dev). The socket at 0o660 owned by root:weaver means only root and weaver group members can connect

internal/daemon/server_test.go
leninmehedy: applied and added test

docs/claude/reviews/00518-implement-unix-socket-server-with-health.md
leninmehedy: updated

internal/daemon/handlers.go
leninmehedy: applied and added test

internal/daemon/server.go
leninmehedy: Made it configurable so that we could test it with shorter delay.

PR #637, opened 2026-06-02
feat(daemon): provision RBAC, kubeconfig and add start/stop commands

leninmehedy:
## Summary

- `solo-provisioner daemon service install` now provisions the full daemon stack end-to-end: checks the K8s cluster, creates the ServiceAccount / ClusterRole / ClusterRoleBinding / token Secret, writes the daemon kubeconfig, then installs and starts the systemd service
- `solo-provisioner daemon service uninstall` tears everything down in reverse
- New `daemon service start` and `daemon service stop` commands for day-2 lifecycle management

## Changes

- `pkg/models/weaver_paths.go` — add `DaemonKubeconfigPath` (`$home/config/daemon.kubeconfig`)
- `internal/workflows/steps/step_daemon_provision.go` — new steps: `CheckClusterStep`, `CreateDaemonRBACStep`, `WriteDaemonKubeconfigStep`, `RemoveDaemonKubeconfigStep`, `DeleteDaemonRBACStep`
- `internal/workflows/daemon.go` — extended install/uninstall workflows; constructors now return `(workflow, error)` to fail fast if `daemon.yaml` is missing
- `cmd/cli/commands/daemon/service/` — updated install/uninstall; new start/stop commands

## Test plan

- `task test:unit` — all unit tests pass
- `task lint` — clean
- Manual UAT: see `docs/claude/reviews/00624-story-daemon-service-provision.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmd/cli/commands/daemon/service/start.go
+++ b/cmd/cli/commands/daemon/service/start.go
@@ -1,0 +1,17 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package service
+
+import (
+	pkgos "github.com/hashgraph/solo-weaver/pkg/os"
+	"github.com/spf13/cobra"
+)
+
+var startCmd = &cobra.Command{
+	Use:   "start",
+	Short: "Start the solo-provisioner-daemon systemd service",
+	Long:  "Start the solo-provisioner-daemon systemd service via systemctl. Requires root privileges.",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return pkgos.StartService(cmd.Context(), daemonServiceName)
+	},
+}
--- a/internal/workflows/daemon.go
+++ b/internal/workflows/daemon.go
@@ -4,30 +4,63 @@ package workflows
 
 import (
 	"github.com/automa-saga/automa"
+	daemon "github.com/hashgraph/solo-weaver/internal/daemon"
 	"github.com/hashgraph/solo-weaver/internal/workflows/steps"
 	"github.com/hashgraph/solo-weaver/pkg/models"
+	"github.com/joomcode/errorx"
 )
 
-// NewDaemonServiceInstallWorkflow installs the solo-provisioner-daemon systemd service unit file
-// into the weaver sandbox, creates a system symlink, enables, and starts the service.
-// Requires root privileges.
-func NewDaemonServiceInstallWorkflow() *automa.WorkflowBuilder {
-	paths := models.Paths()
-	return automa.NewWorkflowBuilder().WithId("daemon-service-install-workflow").Steps(
-		CheckPrivilegesStep(),
-		steps.InstallDaemonServiceStep(paths),
-	)
+// loadOrbit reads the orbit namespace from daemon.yaml. It fails fast if the
+// config is missing or invalid — provisioning cannot proceed without it.
+func loadOrbit(paths models.WeaverPaths) (string, error) {
+	cfg, err := daemon.LoadDaemonConfig(paths.DaemonConfigPath)
+	if err != nil {
+		return "", err
+	}
+	if cfg.Orbit == "" {
+		return "", errorx.IllegalState.New("daemon config %s: orbit namespace must not be empty", paths.DaemonConfigPath)
+	}
+	return cfg.Orbit, nil
 }
 
-// NewDaemonServiceUninstallWorkflow stops, disables, and removes the
-// solo-provisioner-daemon systemd service — both the system symlink and the
-// sandbox unit file. Requires root privileges.
-func NewDaemonServiceUninstallWorkflow() *automa.WorkflowBuilder {
+// NewDaemonServiceInstallWorkflow provisions the full daemon stack:
+//  1. Check root privileges
+//  2. Check K8s cluster is reachable
+//  3. Create RBAC (SA + ClusterRole + CRB + token Secret)
+//  4. Write daemon kubeconfig
+//  5. Install + enable + start systemd service unit
+func NewDaemonServiceInstallWorkflow() (*automa.WorkflowBuilder, error) {
 	paths := models.Paths()
+	orbit, err := loadOrbit(paths)
+	if err != nil {
+		return nil, err
+	}
+	return automa.NewWorkflowBuilder().WithId("daemon-service-install-workflow").Steps(
+		CheckPrivilegesStep(),
+		steps.CheckClusterStep(),
+		steps.CreateDaemonRBACStep(orbit),
+		steps.WriteDaemonKubeconfigStep(paths, orbit),
+		steps.InstallDaemonServiceStep(paths),
+	), nil
+}
+
+// NewDaemonServiceUninstallWorkflow tears down the daemon stack in reverse:
+//  1. Check root privileges
+//  2. Stop + disable + remove systemd service unit
+//  3. Remove daemon kubeconfig
+//  4. Delete RBAC (CRB + CR + Secret + SA)
+func NewDaemonServiceUninstallWorkflow() (*automa.WorkflowBuilder, error) {
+	paths := models.Paths()
+	orbit, err := loadOrbit(paths)
+	if err != nil {
+		return nil, err
+	}
 	return automa.NewWorkflowBuilder().WithId("daemon-service-uninstall-workflow").Steps(
 		CheckPrivilegesStep(),
 		steps.RemoveDaemonServiceStep(paths),
-	)
+		steps.RemoveDaemonKubeconfigStep(paths),
+		steps.DeleteDaemonRBACStep(orbit),
+	), nil
 }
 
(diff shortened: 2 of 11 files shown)


brunodam (commented):
Thanks for putting this together — the end-to-end install/uninstall flow is a really nice improvement, and the `automa.StepBuilder` rollback shape is the right primitive for this. A few things I noticed that I'd love your take on; left most of them inline. Feel free to push back where I'm off base.

### Higher-impact items (not inlined)

- **Unit tests for `step_daemon_provision.go`.** The new file adds ~379 lines of logic (token wait loop, kubeconfig serialization, idempotent create + rollback) without unit tests. `writeDaemonKubeconfig` is pure I/O over a deterministic struct (easy with `t.TempDir()`), and `waitForSAToken` / `CreateDaemonRBACStep` are testable against `fake.NewSimpleClientset()` (already used elsewhere in the codebase). Manual UAT is great for confidence, but it won't catch regressions in CI.
- **`start` / `stop` skip the privilege check.** `install`/`uninstall` start with `CheckPrivilegesStep()`; the new commands call `pkgos.RestartService` / `StopService` directly, so non-root users only see whatever systemctl returns. Adding a privilege preflight (or at least a clear error) would keep the UX consistent.
- **Hard-coded `tokenReadyTimeout = 30s`.** Reasonable on healthy clusters, but a slower control plane could exceed it. Worth either making it configurable or calling it out in the docstring.
- **Multiple kubeconfig loads per workflow run.** `CheckClusterStep`, `CreateDaemonRBACStep`, `WriteDaemonKubeconfigStep`, and `deleteDaemonRBAC` each call `newTypedClient()`, and `waitForSAToken` re-reads the admin kubeconfig on top of that. Not a correctness bug — just a candidate for a small helper that loads once and passes server URL + clientset along.

Nothing here is blocking from my end — happy to chat through any of it. 🙏

cmd/cli/commands/daemon/service/start.go
```diff
@@ -0,0 +1,17 @@
...
+	pkgos "github.com/hashgraph/solo-weaver/pkg/os"
+	"github.com/spf13/cobra"
+)
+
+var startCmd = &cobra.Command{
+	Use:   "start",
+	Short: "Start the solo-provisioner-daemon systemd service",
+	Long:  "Start the solo-provisioner-daemon systemd service via systemctl. Requires root privileges.",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return pkgos.RestartService(cmd.Context(), daemonServiceName)
```
brunodam: Small thing: `start` calls `pkgos.RestartService`, but the `Short`/`Long` say *Start*. In most service managers `start` is a no-op on a running unit, so an operator running this to *ensure availability* probably wouldn't expect an in-flight workload to be torn down and respawned. The review checklist describes this as "idempotent: starts if stopped, restarts if already running" — that's `restart` semantics, not `start`.

Would you be open to either renaming the command to `restart`, or switching to `pkgos.StartService` here? Happy to defer if there's a reason I'm missing.

```suggestion
		return pkgos.StartService(cmd.Context(), daemonServiceName)
```

internal/workflows/daemon.go
```diff
@@ -4,30 +4,59 @@ package workflows
...
-// Requires root privileges.
-func NewDaemonServiceInstallWorkflow() *automa.WorkflowBuilder {
+// loadOrbit reads the orbit namespace from daemon.yaml. It fails fast if the
+// config is missing or invalid — provisioning cannot proceed without it.
+func loadOrbit(paths models.WeaverPaths) (string, error) {
+	cfg, err := daemon.LoadDaemonConfig(paths.DaemonConfigPath)
+	if err != nil {
+		return "", err
+	}
+	return cfg.Orbit, nil
```
brunodam, line 23: The plan doc mentions this should *"fail fast if `daemon.yaml` is missing **or `orbit` is empty**"*, but `loadOrbit` only checks the missing-config case. If `orbit: ""` slips through, the workflow proceeds and we end up creating the SA/Secret in the empty-string namespace — which the API server rejects with a less actionable error than failing here would give.

Might be worth a quick empty-check:

```suggestion
	if cfg.Orbit == "" {
		return "", errorx.IllegalArgument.New("orbit is empty in %s", paths.DaemonConfigPath)
	}
	return cfg.Orbit, nil
```

(Would also need an `errorx` import — happy for you to use whatever error style fits best here.)

internal/workflows/steps/step_daemon_provision.go
```diff
@@ -0,0 +1,379 @@
...
+// RemoveDaemonKubeconfigStep removes the daemon kubeconfig file. Errors for
+// non-existent paths are ignored.
+func RemoveDaemonKubeconfigStep(paths models.WeaverPaths) *automa.StepBuilder {
+	kubeconfigPath := paths.DaemonKubeconfigPath
+	return automa.NewStepBuilder().WithId("remove-daemon-kubeconfig").
+		WithExecute(func(ctx context.Context, stp automa.Step) *automa.Report {
+			if err := os.Remove(kubeconfigPath); err != nil && !os.IsNotExist(err) {
+				logx.As().Warn().Err(err).Str("path", kubeconfigPath).Msg("Failed to remove daemon kubeconfig")
+			}
+			logx.As().Info().Str("path", kubeconfigPath).Msg("Daemon kubeconfig removed")
```
brunodam: On a real `os.Remove` failure (e.g., permission error), this logs **both** *"Failed to remove daemon kubeconfig"* and *"Daemon kubeconfig removed"* and returns success — which is a bit confusing if someone's tailing logs. Same general shape happens in `deleteDaemonRBAC` (line 271 logs *"Daemon RBAC resources deleted"* unconditionally even when individual deletes warned).

One option:

```suggestion
			if err := os.Remove(kubeconfigPath); err != nil {
				if os.IsNotExist(err) {
					logx.As().Info().Str("path", kubeconfigPath).Msg("Daemon kubeconfig already absent")
				} else {
					logx.As().Warn().Err(err).Str("path", kubeconfigPath).Msg("Failed to remove daemon kubeconfig")
				}
			} else {
				logx.As().Info().Str("path", kubeconfigPath).Msg("Daemon kubeconfig removed")
			}
```

No strong opinion on whether to keep the step success-on-error (best-effort cleanup is fine) — just the success-log-on-failure is what stood out.

internal/workflows/steps/step_daemon_provision.go
```diff
@@ -0,0 +1,379 @@
...
+	for {
+		secret, getErr := cs.CoreV1().Secrets(namespace).Get(ctx, daemonTokenSecretName, metav1.GetOptions{})
+		if getErr == nil && len(secret.Data["token"]) > 0 {
+			token = string(secret.Data["token"])
+			ca = string(secret.Data["ca.crt"])
+			break
+		}
+
+		if time.Now().After(deadline) {
+			return "", "", "", fmt.Errorf("SA token secret %s/%s not populated within %s", namespace, daemonTokenSecretName, tokenReadyTimeout)
```
brunodam: Heads up: the accompanying plan doc says *"Error library: `errorx` throughout — no `fmt.Errorf` or `errors.New`."* `waitForSAToken` and `writeDaemonKubeconfig` use `fmt.Errorf` in a few spots (this line, plus 339, 343, 347, 351, and 376). The errors do get wrapped in `errorx.InternalError.Wrap(...)` by the callers, so it's not a correctness issue — just a small inconsistency with what the doc promises. Either swapping these to `errorx` or relaxing the doc would clear it up. 🙂

internal/workflows/steps/step_daemon_provision.go
```diff
@@ -0,0 +1,379 @@
...
+				Str("secret", daemonTokenSecretName).
+				Msg("Daemon RBAC resources created")
+			return automa.StepSuccessReport(stp.Id())
+		}).
+		WithPrepare(func(ctx context.Context, stp automa.Step) (context.Context, error) {
+			notify.As().StepStart(ctx, stp, "Creating daemon RBAC resources")
+			return ctx, nil
+		}).
+		WithRollback(func(ctx context.Context, stp automa.Step) *automa.Report {
+			deleteDaemonRBAC(ctx, namespace)
```
brunodam: Worth thinking about: `CreateDaemonRBACStep` is (intentionally) idempotent — it treats `AlreadyExists` as success — but the `WithRollback` here unconditionally calls `deleteDaemonRBAC`. If a previous install left the SA/CR/CRB in place and a *later* step in this run fails, rollback will delete resources we didn't actually create on this attempt, including the `solo-provisioner-daemon-token` Secret that an existing kubeconfig on disk may still depend on.

Few ways to handle it: track which resources were actually created in this run (e.g., set a bool when `Create` succeeds without `AlreadyExists`) and only roll those back, or document that rollback is intentionally destructive. Not a blocker — flagging because it's the kind of thing that bites later when someone re-runs install.

Issue #562, opened 2026-05-14
Story 9.2 — Validate declared infrastructure versions against the built-in catalog


leninmehedy:
**Epic:** #506 — Infrastructure Version Management

Reject any deployment package whose `infrastructure-versions.yaml` declares versions outside the supported combination for the current `provisioner.cli.version`.

**Acceptance criteria:**
- Validation runs on package processing, before any upgrade step executes
- If any declared version is outside the catalog entry, reject with a clear error message naming the offending version
- Emit a K8s warning event on rejection

Issue #554, opened 2026-05-14
Story 7.3 — Expose state download as solo-provisioner consensus node restore command


leninmehedy:
**Epic:** #504 — state-sources.yaml Download

Wire the state snapshot download into a user-facing CLI command for fresh node provisioning and disaster recovery.

**Acceptance criteria:**
- Command: `solo-provisioner consensus node restore` (or equivalent)
- Reads `state-sources.yaml` from the deployment package
- Runs full download-and-verify flow (Stories 7.1 + 7.2)
- Provides progress output and clear success/failure messages

Issue #541, opened 2026-05-14
Story 5.5 — Implement ConsensusConfig CR creation from data/config/ scan


leninmehedy:
**Epic:** #502 — Network Upgrade Workflow (Execute Phase)

Scan the `data/config/` directory in the deployment package and create one dedicated ConsensusConfig CR
per recognised config file.

**Acceptance criteria:**
- Use a hardcoded filename-to-CR-kind mapping (e.g. `log4j2.xml` → `Log4j2Config`, `settings.txt` → `NodeSettings`, `throttles.json` → `ThrottleConfig`). The full mapping + CR spec/GVR/label/status contract comes from solo-operator (see blocker).
- Determine the capsule name by querying the orbit namespace for the `ConsensusCapsule` CR and matching on the node-id label; use it for `metadata.name` and the `hedera.com/capsule` label. `spec.orbit` comes from the Execute CR.
- Reject unrecognised filenames with a `ConfigCRFailed` event; do not silently skip.
- Apply the `hedera.com/operation-id` label to every CR created.
- Emit `ConfigCRCreated` per CR.

**Blocked on:** solo-operator CRD contract (finalized filename→kind table, GVRs, spec/label/status field paths).

Issue #540, opened 2026-05-14
Story 5.4 — Implement infra upgrade detection and orchestration


leninmehedy:
**Epic:** #502 — Network Upgrade Workflow (Execute Phase)

**This story has been split into three** for clear sequencing and separate testability:

- **5.4a — Infra-upgrade detection** (read software_state, compare host/cluster, set `PendingInfraUpgrade`).
- **5.4b — Cluster rebuild via forked CLI** (hand off to epic #500's detached child; phase stays `PendingInfraUpgrade`). _Blocked on #500._
- **5.4c — Resume from `PendingInfraUpgrade`** (teach `listAndSeed` to resume after daemon restart; post-rebuild RBAC backoff). _Blocked on #500._

Child sub-issues: #707 (5.4a), #708 (5.4b), #709 (5.4c). This umbrella issue is retained for traceability; implementation happens in the three child stories.

PR #642, opened 2026-06-03
feat(manifests): parse and validate CN deployment-package YAML files

### LINKED ISSUE #501: Epic A38_4 — Manifest Parsing and Validation
Parse and validate all `manifests/` YAML files from the deployment package.

### LINKED ISSUE #531: Story: Implement consensus-node-components.yaml schema v1 parser
Part of epic #501 — Manifest Parsing and Validation.

Parser must handle: named sidecar components (`recordStreamUploader`, `eventStreamUploader`, `blockStreamUploader`, `backupUploader`, `uc`), `enabled` field, optional sections (absent = no change), deterministic and non-deterministic `layerHashes`.

### LINKED ISSUE #532: Story: Implement infrastructure-versions.yaml parser with provisioner.cli and provisioner.daemon sections
Part of epic #501 — Manifest Parsing and Validation.

Parser must handle: `provisioner.cli` and `provisioner.daemon` each with `version`, `algorithm`, `checksum`; `host` section (CRI-O, kubelet, host binaries); `cluster` section (Helm charts and components).

### LINKED ISSUE #533: Story: Implement external-files.yaml parser with algorithm+checksum, optional flag, and phase validation
Part of epic #501 — Manifest Parsing and Validation.

Parser must handle: `algorithm` + `checksum` fields, `optional` flag (default false), `phase.download` (`prepare` or `freeze`), `phase.install` (`freeze`), and directory marker prefix validation on `destination`.

### LINKED ISSUE #534: Story: Implement state-sources.yaml parser for multi-bucket GCS/S3 sources with index and paths
Part of epic #501 — Manifest Parsing and Validation.

Parser must handle: multiple bucket entries (GCS, S3), per-bucket `type`, `location`, `index` map (node → index file path), and `paths` map (node → base path for state snapshots).

### LINKED ISSUE #535: Story: Implement schemaVersion validation across all manifest files
Part of epic #501 — Manifest Parsing and Validation.

Validate `schemaVersion` field on every manifest file on load. Reject unknown schema versions with a clear error before any further processing.

### LINKED ISSUE #536: Story: Enforce directory marker prefix restriction on external-files.yaml destination field
Part of epic #501 — Manifest Parsing and Validation.

Enforce that the `destination` field in `external-files.yaml` uses only recognised marker prefixes (`HAPIAPP_DIR`, `SOLO_PROVISIONER_DIR`). Reject arbitrary filesystem paths with a clear error.

brunodam:
## Summary

Epic #501 introduces `pkg/manifests/` — parsers and validators for the four YAML files shipped inside every CN deployment package: `consensus-node-components.yaml`, `infrastructure-versions.yaml`, `external-files.yaml`, and `state-sources.yaml`. The epic is labelled `order: foundation` because downstream epics (notably #502 — Network Upgrade Workflow Execute Phase) cannot consume the deployment package until these parsers exist.

This is the **umbrella PR for the whole epic**. The six story PRs (#634, #636, #638, #639, #640, #641) remain open as the granular per-story review surface; this PR is the integrator's view, showing the cumulative diff against `main` so reviewers and release engineering can see the full epic in one place. On merge, GitHub will auto-close the six story PRs once their commit SHAs land on `main`.

### Stacked stories (all included)

| Story | Title | PR |
|---|---|---|
| #535 | schemaVersion validator (cross-cutting) | #634 |
| #531 | `consensus-node-components.yaml` parser | #636 |
| #532 | `infrastructure-versions.yaml` parser | #638 |
| #533 | `external-files.yaml` parser | #639 |
| #534 | `state-sources.yaml` parser | #640 |
| #536 | Destination marker-prefix enforcement | #641 |

Plus one epic-level commit adding [`docs/claude/reviews/00501-manifest-parsing.md`](docs/claude/reviews/00501-manifest-parsing.md).

### Merge strategy

Use **"Create a merge commit"** or **"Rebase and merge"**, not "Squash". Squashing would change the commit SHAs and prevent GitHub from auto-closing the six story PRs.

### Scope boundary

No production caller — parsers are not yet imported anywhere outside `pkg/manifests/`. Wiring lives in epic #502 (stories #537–#541) and two smaller integrations in epics #504 (#554) and #506 (#562). See the demo doc §5 for the full table.

### Review guide

# #501 — Demo guide: Epic A38_4, Manifest Parsing and Validation

> **Epic:** https://github.com/hashgraph/solo-weaver/issues/501
> **Feature branch:** `00501-manifest-parsing` (off `main`)
> **Tip of stack:** `00536-destination-marker-prefix` (PR #641) — transitively contains every story; use this checkout to demo
> **Stories:** #531, #532, #533, #534, #535, #536 — per-story review docs are linked at the bottom

## What this epic delivers

A new `pkg/manifests/` package that parses and validates the four YAML files shipped inside every CN deployment package:

| File | Purpose | Story |
|---|---|---|
| `consensus-node-components.yaml` | Container images + layer-hash integrity for the CN and its five sidecars | #531 |
| `infrastructure-versions.yaml` | Pinned `provisioner.cli`/`provisioner.daemon` versions + audit records for every host binary / Helm chart | #532 |
| `external-files.yaml` | Large remote files (>1 MB) the upgrade-controller / `solo-provisioner-upgrade` daemon downloads and stages | #533, #536 |
| `state-sources.yaml` | GCS / S3 buckets for fast-sync of saved-state snapshots | #534 |

Cross-cutting under the same package:

- `ValidateSchemaVersion(kind, data)` — per-kind allowlist of schema versions, runs before any structural decode (#535).
- `decodeStrictSingleYAMLDoc(kind, data, out)` — strict YAML decode helper (`KnownFields(true)` + multi-document rejection) used by all four parsers.
- Typed `errorx` errors: `ParseError`, `MissingSchemaVersionError`, `UnsupportedSchemaVersionError`, `UnknownKindError`, `ValidationError` — every failure mode is `errorx.IsOfType`-classifiable.
- Field-path errors with dotted Go paths (e.g. `images.backupUploader.registries[0].layerHashes`) — greppable against the YAML source.

## What this epic intentionally does not deliver

- **No production caller.** The parsers exist but nothing in the daemon or CLI imports them yet. Wiring into the package-extraction flow / fast-sync bootstrap / UC apply is a follow-on epic.
- **No CLI surface.** Nothing the operator can run by hand — exercise is exclusively unit-test-driven.
- **No catalog cross-checks.** `infrastructure-versions.yaml` host[]/cluster[] names are not validated against `pkg/software/infrastructure-catalog.yaml`; that's a runtime concern, not a parse concern.

Frame this explicitly during the demo: this is the **foundation layer** that downstream epics will plug into. If the audience expects to see a feature, that's the wrong expectation.

### 1. Frame the epic (1 min)
Open issue #501. Read the one-liner ("Parse and validate all `manifests/` YAML files from the deployment package") and call out the `order: foundation` label — others can't start until this lands.

### 2. API tour (4 min)
From the tip checkout, walk `pkg/manifests/` top-down in dependency order:

```
manifests.go              # Kind enum, SchemaVersion, supportedVersions registry, ValidateSchemaVersion
decode.go                 # decodeStrictSingleYAMLDoc helper
errors.go                 # errorx namespace + 5 typed errors
consensus_node_components.go
infrastructure_versions.go
external_files.go         # includes AllowedDestinationPrefixes() — cloned slice, callers can't weaken enforcement
state_sources.go          # includes AllowedBucketSchemes() — provider type is encoded in the bucket URI scheme (gcs://, s3://)
```

Each parser has the same shape: `Parse<Kind>(data)` → `ValidateSchemaVersion` → strict decode → semantic validate. The repetition is the point — adding a fifth YAML kind is a one-file change.

### 3. Validation in action (5 min)
Run the unit suite verbosely and let the test names do the narration:

```bash
go test -v -tags='!integration' ./pkg/manifests/... 2>&1 | head -120
```

Then surface four representative failure modes by name (one per parser):

```bash
# schemaVersion gate (cross-cutting)
go test -v -tags='!integration' -run 'UnsupportedSchemaVersion' ./pkg/manifests/...

# Deterministic vs registry-level layerHashes placement (#531)
go test -v -tags='!integration' -run 'DeterministicLayerHashesPlacement|NonDeterministicLayerHashesPlacement' ./pkg/manifests/...

# Duplicate destination across external-files entries (#533)
go test -v -tags='!integration' -run 'DuplicateDestination' ./pkg/manifests/...

# Marker-prefix allowlist on destination (#536)
go test -v -tags='!integration' -run 'DestinationPrefixEnforcement' ./pkg/manifests/...
```

For each, point out the field-path in the error message — that's the operator-facing contract.

### 4. HIP alignment (2 min)
HIP-XXXX0 has caught up to the story bodies on the earlier shape divergences (named sidecars on consensus-node-components, split `provisioner.cli`/`provisioner.daemon`, separate `algorithm`+`checksum` on external-files). Regression tests remain as guards against the old shapes resurfacing — show them as a single grep:

```bash
grep -nE "RejectsHIP|RejectsUnknownComponentName" pkg/manifests/*_test.go
```

Two HIP-driven alignment changes landed during PR review (after the six per-story PRs were stacked):

| Change | Driver | Pinned by |
|---|---|---|
| `schemaVersion` is now an `int` (HIP says `schemaVersion: 1`, not `"v1"`) | DevOps wanted to avoid string→int parsing | `TestValidateSchemaVersion_NonIntegerValue`, `TestValidateSchemaVersion_AcceptsV1ForEveryKnownKind` |
| `state-sources.yaml` drops the `type` field; provider is encoded in the bucket URI scheme (`gcs://`, `s3://`) | HIP-XXXX0 §"state-sources.yaml": *"removes the need for a separate `type` field (the scheme encodes the provider)"* | `TestParseStateSources_ValidationFailures` (unsupported-scheme, HTTPS-style, scheme-alone cases), `TestAllowedBucketSchemes_PinnedSet` |

### 5. Scope boundary and what's next (2 min)
- Parsers compile but aren't called. Show `grep -rn 'pkg/manifests' --include='*.go' | grep -v _test.go | grep -v 'pkg/manifests'` — empty.
- The integration lands primarily in **Epic #502 — A38_5 Network Upgrade Workflow (Execute Phase)** (`order: core`, depends on this epic). Five of its eight stories directly consume the #501 parsers:

  | Story | Parser consumed | What it wires up |
  |---|---|---|
  | #537 (5.1) | `external-files.yaml` (#533/#536) | Download step + disk-space check + halt/skip policy |
  | #538 (5.2) | `external-files.yaml` (#533/#536) | Atomic-write install during freeze; honours `HAPIAPP_DIR`/`SOLO_PROVISIONER_DIR` markers |
  | #539 (5.3) | `infrastructure-versions.yaml` (#532) | Atomic-write placement before infra upgrade |
  | #540 (5.4) | `infrastructure-versions.yaml` (#532) | Detect declared-vs-installed delta, orchestrate teardown/reinstall |
  | #541 (5.5) | `consensus-node-components.yaml` (#531) | ConsensusConfig CR creation from `data/config/` scan |

- Two smaller integrations live outside #502:
  - **Epic #504 → Story #554** (7.3) — `solo-provisioner consensus node restore` CLI command consumes `state-sources.yaml` (#534).
  - **Epic #506 → Story #562** (9.2) — validate declared infra versions against the built-in catalog (consumes #532). This is also where the catalog cross-check that we deliberately deferred from #532 lands.

- During the demo, the most quotable single forward-pointer is **#540** ("Detect when infrastructure versions declared in the manifest differ from what is installed") — its acceptance criteria read like a one-line user story for why this epic exists.

## Demo prep checklist

- [ ] `git fetch origin && git checkout 00536-destination-marker-prefix && git pull` — tip of the stack has everything.
- [ ] `go test -race -cover -tags='!integration' ./pkg/manifests/...` shows ~97.9% coverage, no failures.
- [ ] `task lint:check` is clean.
- [ ] Stack visualisation ready: epic ← #634 (#535) ← #636 (#531) ← #638 (#532) ← #639 (#533) ← #640 (#534) ← #641 (#536). None of the six has merged at demo time; that's expected — they unroll into the epic branch in order, then the epic merges to `main`.

## Per-story review docs

Each story ships its own review guide; reference these for deep-dives during Q&A.

| Story | Title | PR | Doc |
|---|---|---|---|
| #535 | schemaVersion validator (cross-cutting) | #634 | `docs/claude/reviews/00535-manifest-schema-version-validation.md` |
| #531 | `consensus-node-components.yaml` parser | #636 | `docs/claude/reviews/00531-consensus-node-components-parser.md` |
| #532 | `infrastructure-versions.yaml` parser | #638 | `docs/claude/reviews/00532-infrastructure-versions-parser.md` |
| #533 | `external-files.yaml` parser | #639 | `docs/claude/reviews/00533-external-files-parser.md` |
| #534 | `state-sources.yaml` parser | #640 | `docs/claude/reviews/00534-state-sources-parser.md` |
| #536 | destination marker-prefix enforcement | #641 | `docs/claude/reviews/00536-destination-marker-prefix.md` |

## Risks and open questions to surface during Q&A

- **HIP alignment is the live tracking surface.** The HIP and the parsers now agree on `schemaVersion: int` and on dropping the `type` field from `state-sources.yaml`. The remaining regression tests (`RejectsHIP…`) guard against earlier shapes resurfacing in fixtures, not against unresolved disagreement.
- **Algorithm field is not enum-restricted.** `Binary.algorithm` accepts any non-empty string; a future "must be sha256" gate would live centrally in `pkg/software/config.go::Checksum`, not in the parsers.
- **`Optional` (on `external-files` entries) is a bool, not `*bool`.** Distinguishing nil from explicit-false would need a pointer — out of scope for this epic; flag if downstream needs it.
- **No catalog membership check** on `infrastructure-versions.yaml` host[]/cluster[] names. Deliberate scope boundary — surface the choice so reviewers don't expect it here.

#### Story 00531

# #531 — Review guide: consensus-node-components.yaml parser

> **Issue:** https://github.com/hashgraph/solo-weaver/issues/531
> **Epic:** https://github.com/hashgraph/solo-weaver/issues/501
> **Branch:** `00531-consensus-node-components-parser`
> **Base:** `00535-manifest-schema-version-validation` (stacked; auto-retargets to `00501-manifest-parsing` after #535 merges)

## Problem and solution

`consensus-node-components.yaml` ships inside every CN deployment package and declares the container images (consensus node plus five named sidecars: `recordStreamUploader`, `eventStreamUploader`, `blockStreamUploader`, `backupUploader`, `uc`) along with the layer-hash integrity records used to verify those images before they run. This story is the first concrete parser story under epic #501 and the first consumer of `ValidateSchemaVersion` from PR #634.

The parser strict-decodes the YAML into `ConsensusNodeComponents`, then runs semantic validation that goes beyond what struct shape alone can enforce — most importantly the *placement* rule for `layerHashes`: deterministic builds carry one shared set at the component level and no per-registry overrides; non-deterministic builds invert that. Getting this wrong silently would mean a missing or duplicated integrity record at apply time, so the validator rejects every misplacement explicitly.

### Spec note

The story body lists the five sidecar names above; the older HIP draft (`hip-xxxx0`) shows a single bundled `cheetah` sidecar. Stories supersede the HIP draft (same convention as the `provisioner.cli`/`provisioner.daemon` split flagged in #535's plan). Strict decoding therefore rejects `cheetah` (or any other unknown component name) — `TestParseConsensusNodeComponents_RejectsUnknownComponentName` pins that.

### "Absent = no change"

Per the manifest contract, every entry under `images:` is optional. A nil pointer (Go-side) corresponds to "section absent from the YAML" and signals "no change to that component". A separate `enabled *bool` field per entry encodes the operator's intent to turn a component on or off; nil there means "no change to the current enabled state" — distinct from `enabled: false` which is an explicit disable.

## Changed files

| File | What changed |
|---|---|
| `pkg/manifests/consensus_node_components.go` (new) | Types (`ConsensusNodeComponents`, `Images`, `Image`, `Deterministic`, `Registry`, `LayerHashes`), `SupportedImagePlatforms`, `ParseConsensusNodeComponents`, internal validators |
| `pkg/manifests/consensus_node_components_test.go` (new) | Happy paths (full, partial, empty `images:`), strict-decode rejections, schemaVersion gate, table-driven semantic-validation failures, multi-document rejection |
| `pkg/manifests/decode.go` (new) | Shared `decodeStrictSingleYAMLDoc(kind, data, out)` helper used by every concrete parser: enables `KnownFields(true)` and rejects multi-document inputs |
| `pkg/manifests/errors.go` | Added `ValidationError` type and `NewValidationError(kind, field, reason)` constructor; added `fieldProperty` printable property |
| `docs/claude/reviews/00531-consensus-node-components-parser.md` (new) | This guide |

## Code review checklist

- [ ] **`Image` field types model the "absent = no change" contract.** `Enabled *bool` distinguishes nil (no change), `true` (enable), and `false` (disable). Compare against `Deterministic *Deterministic` and the per-component `*Image` fields under `Images` — the pointer convention is consistent throughout.
- [ ] **`ParseConsensusNodeComponents` runs `ValidateSchemaVersion` first.** A future-versioned manifest is rejected before any current-shape decode can yield misleading errors. `TestParseConsensusNodeComponents_RejectsUnsupportedSchemaVersion` pins it.
- [ ] **Strict decode is enabled at this layer** (via the new `decodeStrictSingleYAMLDoc` helper in `pkg/manifests/decode.go`, which sets `dec.KnownFields(true)` and rejects multi-document inputs). Unknown top-level fields and unknown component names under `images:` both fail with `ParseError`. Multi-document YAML streams fail with `ValidationError` — see `TestParseConsensusNodeComponents_RejectsMultipleYAMLDocuments`. The `KnownFields(true)` strictness at parser level is intentional contrast with the schemaVersion validator from #535, which tolerates unknown fields by design (`TestValidateSchemaVersion_ToleratesUnknownFields` in PR #634).
- [ ] **Deterministic vs non-deterministic placement is enforced both ways.** Six dedicated test cases cover all four combinations (`deterministic.supported` × `layerHashes` present at component-level vs registry-level). The validation error messages explain which direction to move the data.
- [ ] **Platform identifiers are validated against `SupportedImagePlatforms`** (`linux/amd64`, `linux/arm64`). Unknown platform keys produce an error naming the offending value; empty hash lists per platform are rejected (a platform key with `[]` is meaningless).
- [ ] **Field paths in `ValidationError` messages are dotted Go paths** (e.g. `images.backupUploader.registries[0].layerHashes`). The format makes errors greppable against the YAML source and makes review failures easy to locate.
- [ ] **No production caller yet.** This PR ships the parser; #531 says nothing about wiring it into a workflow. That's deferred — the parser is exercised exclusively by its unit tests until a daemon/operator integration story picks it up.

## Test commands

```bash
# Unit tests for the parser (and the existing schemaVersion validator)
go test -race -cover -tags='!integration' ./pkg/manifests/...

# Lint pipeline (errorx gate, gofmt)
task lint:check
```

Expected:

```
ok      github.com/hashgraph/solo-weaver/pkg/manifests    coverage: 97.7% of statements
```

## Manual UAT

No CLI surface, no daemon hook, no production caller yet. Reviewers can spot-check the parser shape against the spec by running this REPL-style snippet in the repo root:

```bash
cat <<'EOF' | go run -trimpath -ldflags='-s -w' ./pkg/manifests/... 2>/dev/null || true
# (no main package — read the fixture instead)
EOF
cat pkg/manifests/consensus_node_components_test.go | sed -n '/fullDeterministicManifest = /,/^$/p' | head
```

The intended user-facing experience surfaces once a follow-on story (outside epic #501) wires this parser into the daemon's package-extraction flow.

#### Story 00532

# #532 — Review guide: infrastructure-versions.yaml parser

> **Issue:** https://github.com/hashgraph/solo-weaver/issues/532
> **Epic:** https://github.com/hashgraph/solo-weaver/issues/501
> **Branch:** `00532-infrastructure-versions-parser`
> **Base:** `00531-consensus-node-components-parser` (stacked; auto-retargets up the chain as upstream PRs merge)

## Problem and solution

`infrastructure-versions.yaml` ships inside every CN deployment package and pins two things the provisioner needs at apply time:

1. **The exact `solo-provisioner` versions** to install — split into `provisioner.cli` (CLI binary) and `provisioner.daemon` (daemon binary), each with `version` + `algorithm` + `checksum`. The split reflects the existing two-binary release layout from epic #498 (CLI tags `vX.Y.Z`; daemon tags `daemon-vX.Y.Z` — independently released).
2. **An audit record** of every host-level binary (`cri-o`, `kubelet`, `kubeadm`, etc.) and Helm chart (`alloy`, `metallb`, `metrics-server`, etc.) whose installed version the provisioner will reconcile against its embedded catalog at runtime. These are deliberately **audit-only** records — they do not select versions; `provisioner.<cli|daemon>.version` does.

The parser strict-decodes the YAML, then validates: provisioner sub-sections have all three integrity fields when present, host[] and cluster[] entries have non-empty name + version, and names are unique within each section (duplicate names would produce ambiguous audit records).

### Spec note — supersedes the HIP draft

The HIP draft (`hip-xxxx0`) shows a single `provisioner.version` field. The story body for #532 supersedes that with the `cli`/`daemon` split — same convention as the `cheetah` → 5-sidecar split flagged in #531. Strict decoding therefore rejects a manifest with `provisioner.version` at the top level of `provisioner:` — `TestParseInfrastructureVersions_RejectsHIPSingleProvisionerVersion` pins this contract.

### Stacked-PR layout

This branch is stacked on `00531-consensus-node-components-parser` (PR #636), which is itself stacked on `00535-manifest-schema-version-validation` (PR #634). The reason is the shared `ValidationError` errorx type and `NewValidationError` constructor that #531 added to `pkg/manifests/errors.go` — both #532 and #531 use it. Stacking avoids a merge conflict on `errors.go` and keeps the typed-error surface defined exactly once.

As the upstream PRs merge into the epic branch, GitHub will auto-retarget this PR down the chain (PR #636 first, then this one), arriving at `00501-manifest-parsing` cleanly.

## Changed files

| File | What changed |
|---|---|
| `pkg/manifests/infrastructure_versions.go` (new) | Types (`InfrastructureVersions`, `Provisioner`, `Binary`, `HostComponent`, `ClusterChart`), `ParseInfrastructureVersions`, internal validators |
| `pkg/manifests/infrastructure_versions_test.go` (new) | Happy paths (full + partial), schemaVersion gate, strict-decode rejections (incl. the HIP-shape rejection), table-driven validation failures including duplicate-name cases |
| `docs/claude/reviews/00532-infrastructure-versions-parser.md` (new) | This guide |

## Code review checklist

- [ ] **`Provisioner`, `Provisioner.CLI`, `Provisioner.Daemon` are pointer-typed.** A manifest with only `provisioner.cli` present (or with no `provisioner:` block at all) is valid under "absent = no change" — verified by `TestParseInfrastructureVersions_OnlyCLIPresent` and `TestParseInfrastructureVersions_AllSectionsAbsent`.
- [ ] **Strict decode at this layer.** Unknown top-level fields and unknown sub-keys under `provisioner:` (most importantly the old HIP-shape `provisioner.version`) fail with `ParseError`. Two dedicated test cases pin this.
- [ ] **All three integrity fields required when a `Binary` is present.** `Binary.validate` checks `version`, `algorithm`, and `checksum` individually so a missing one produces a precise field-path error.
- [ ] **Host and cluster name uniqueness is per-section.** Same name in both `host[]` and `cluster[]` is allowed (`TestParseInfrastructureVersions_NamesShareableAcrossSections`) — but duplicates *within* a section are rejected.
- [ ] **Algorithm is not enum-restricted.** The Binary struct accepts any non-empty algorithm string, matching the convention in `pkg/software/config.go::Checksum`. A future "must be sha256" enforcement can be added centrally there once the catalog standardises on it.
- [ ] **Validator does not depend on host[]/cluster[] catalog membership.** The provisioner catalog membership check (whether the named host or cluster component is actually recognised by the embedded `pkg/software/infrastructure-catalog.yaml`) is intentionally a runtime concern, not a manifest-parse concern — keeping this PR focused on syntax + per-entry semantics.
- [ ] **No production caller yet.** Like #531, this story ships the parser only. Wiring into the daemon/UC apply flow is a follow-on outside epic #501.

## Test commands

```bash
# Unit tests (this PR brings the package to 98.5% coverage)
go test -race -cover -tags='!integration' ./pkg/manifests/...

# Lint pipeline (errorx gate, gofmt)
task lint:check
```

Expected:

```
ok      github.com/hashgraph/solo-weaver/pkg/manifests    coverage: 98.5% of statements
```

## Manual UAT

No CLI surface, no daemon hook. Reviewers can spot-check the parser shape by reading the fixture at the top of `pkg/manifests/infrastructure_versions_test.go` (`fullInfrastructureVersionsManifest`) against `pkg/manifests/infrastructure_versions.go::InfrastructureVersions`. The intended user-facing behaviour surfaces once a follow-on story wires this parser into the daemon's package-extraction flow.

#### Story 00533

# #533 — Review guide: external-files.yaml parser

> **Issue:** https://github.com/hashgraph/solo-weaver/issues/533
> **Epic:** https://github.com/hashgraph/solo-weaver/issues/501
> **Branch:** `00533-external-files-parser`
> **Base:** `00532-infrastructure-versions-parser` (stacked; auto-retargets up the chain as upstream PRs merge)

## Problem and solution

`external-files.yaml` declares large remote files (over the ~1 MB ConfigurationFile-CR limit) that the upgrade-controller sidecar or the `solo-provisioner-upgrade` daemon must download and stage on the host before the consensus node restarts. Each entry specifies where to fetch the file from, how to verify its integrity (`algorithm` + `checksum`), where to put it (`destination`), whether a download failure is fatal (`optional`), and the timing of both download and install (`phase`).

The parser strict-decodes the YAML and validates: every entry has a non-empty `url`/`algorithm`/`checksum`/`destination`; `phase.download` is one of `{prepare, freeze}`; `phase.install` is `freeze` (the only legal value today). Destination uniqueness across the file list is enforced — two entries can't install to the same path, since they would silently overwrite each other.

### Spec note — supersedes the HIP draft

The HIP draft (`hip-xxxx0`) shows a single `hash: "sha256:..."` field; the story body for #533 supersedes that with separate `algorithm` + `checksum` fields (matching the convention already used in `pkg/software/config.go::Checksum` and now in `infrastructure-versions.yaml`'s `Binary` from #638). A manifest using the old shape fails parsing — `TestParseExternalFiles_RejectsHIPHashFieldShape` pins this.

### Scope boundary with #536

Story #533 implements the parser *infrastructure* for `destination`, validating that the field is non-empty. Story **#536** layers on the strict allowlist of marker prefixes (`HAPIAPP_DIR`, `SOLO_PROVISIONER_DIR`) — that's intentionally a separate PR so the policy is easy to audit and revisit independently of the parser. The test `TestParseExternalFiles_DestinationPrefixNotYetEnforced` pins this scope split: an arbitrary absolute path passes here, and #536 will be the PR that flips that.

### Stacked-PR layout

This branch is stacked on `00532-infrastructure-versions-parser` (PR #638), which is stacked on `00531-…` (PR #636), which is stacked on `00535-…` (PR #634). All four PRs target the epic branch `00501-manifest-parsing`. As upstream PRs merge, GitHub will auto-retarget each downstream PR down the chain.

## Changed files

| File | What changed |
|---|---|
| `pkg/manifests/external_files.go` (new) | Types (`ExternalFiles`, `ExternalFile`, `Phase`, `DownloadPhase`, `InstallPhase`), `ParseExternalFiles`, per-entry validation |
| `pkg/manifests/external_files_test.go` (new) | Happy path, empty-files no-op, `optional`-defaults-false pin, strict-decode rejections (incl. HIP-shape), table-driven validation failures including duplicate-destination, `phase.download`/`phase.install` enum cases, and the destination-prefix-deferred-to-#536 pin |
| `docs/claude/reviews/00533-external-files-parser.md` (new) | This guide |

## Code review checklist

- [ ] **`DownloadPhase` and `InstallPhase` are typed constants.** Validation switches over the typed constants so misspellings in production callers (e.g. `manifests.DownloadPhasePrepar`) fail to compile rather than fail at runtime.
- [ ] **`Optional` is a bool (not a pointer).** The story specifies a default of `false`; `yaml.v3` zeroes the field when absent, which is the same as the documented default. Pinned by `TestParseExternalFiles_OptionalDefaultsFalse`. Future enhancement (distinguishing nil from explicit-false) would need a pointer, but that's not in scope here.
- [ ] **Strict decode at this layer.** Unknown top-level fields and unknown per-entry fields (most importantly the HIP-shape `hash:`) fail with `ParseError`. Two dedicated test cases pin this.
- [ ] **Duplicate-destination check uses a map keyed by destination string.** The previous index is included in the error message so the operator knows both lines to look at.
- [ ] **Empty files list is valid.** `TestParseExternalFiles_EmptyFilesList` pins that a manifest with only `schemaVersion: 1` is a structurally valid no-op — useful in environments where no external files exist for a given release.
- [ ] **Scope boundary with #536.** `TestParseExternalFiles_DestinationPrefixNotYetEnforced` is the *contract* boundary: this PR validates destination is non-empty, #536 will add the strict allowlist. The test will be replaced (not deleted) in #536 with assertions that the strict allowlist now applies — clean, audit-able transition.

## Test commands

```bash
# Unit tests (this PR brings the package to 98.8% coverage)
go test -race -cover -tags='!integration' ./pkg/manifests/...

# Lint pipeline (errorx gate, gofmt)
task lint:check
```

Expected:

```
ok      github.com/hashgraph/solo-weaver/pkg/manifests    coverage: 98.8% of statements
```

## Manual UAT

No CLI surface, no daemon hook. The fixture at the top of `pkg/manifests/external_files_test.go` (`fullExternalFilesManifest`) is the canonical reference shape — reviewers can read it alongside the type definitions in `pkg/manifests/external_files.go`. The intended user-facing behaviour surfaces once a follow-on story wires this parser into the daemon/UC apply flow.

#### Story 00534

# #534 — Review guide: state-sources.yaml parser

> **Issue:** https://github.com/hashgraph/solo-weaver/issues/534
> **Epic:** https://github.com/hashgraph/solo-weaver/issues/501
> **Branch:** `00534-state-sources-parser`
> **Base:** `00533-external-files-parser` (stacked; auto-retargets up the chain as upstream PRs merge)

## Problem and solution

`state-sources.yaml` declares the cloud storage locations from which a new or rejoining consensus node can fast-sync the latest saved-state snapshot — avoiding a full replay of the event stream from genesis. Multiple buckets across GCS, S3, and different regions are listed for redundancy and geographic locality. For each bucket, two parallel maps keyed by node ID locate the per-node index file (which names the latest available round) and the per-node base path (under which the round's state files live).

The parser strict-decodes the YAML and validates per source: `bucket` non-empty and prefixed with a recognised cloud-storage scheme (`gcs://`, `s3://`); `location` non-empty; `index` and `paths` both non-empty maps; *and* their key sets must match exactly (an index pointing at a node with no path is unfollowable; a path with no index has no latest-round pointer). Bucket URIs are unique across sources — two entries pointing at the same bucket would be ambiguous. The provider type is encoded in the bucket URI scheme rather than carried as a separate `type` field, per HIP-XXXX0.

### Stacked-PR layout

At the top of the four-parser stack: `00501-…` ← #634 (#535) ← #636 (#531) ← #638 (#532) ← #639 (#533) ← **this PR**. All five target the epic branch via the chain; GitHub auto-retargets each downstream PR as upstreams merge.

This PR is the first concrete consumer of the `decodeStrictSingleYAMLDoc` helper introduced in #636's amend pass — written from the start to use it (the three earlier parsers were retrofitted via the cascade).

## Changed files

| File | What changed |
|---|---|
| `pkg/manifests/state_sources.go` (new) | Types (`StateSources`, `StateSource`), `ParseStateSources`, `AllowedBucketSchemes()`, per-source validation including bucket-scheme enforcement, index/paths key-set consistency, and cross-source bucket uniqueness |
| `pkg/manifests/state_sources_test.go` (new) | Happy path, empty-stateSources no-op, schemaVersion gate (including a top-level `version:` rejection), strict-decode + multi-doc rejection, table-driven validation-failure cases covering unsupported schemes and HTTPS-style URLs, same-node-across-buckets-allowed pin, and `AllowedBucketSchemes` clone/pin tests |
| `docs/claude/reviews/00534-state-sources-parser.md` (new) | This guide |

## Code review checklist

- [ ] **Provider type is encoded in the bucket URI scheme.** The HIP removes the separate `type` field; `gcs://` and `s3://` prefixes are the source of truth for picking the SDK client. `validateBucketScheme` enforces this and rejects HTTPS-style URLs, unknown schemes, and a bare scheme with no bucket name. The error message lists the supported set.
- [ ] **`AllowedBucketSchemes()` is exported and returns a clone.** External tooling can reuse the closed set without forking it; the unexported `allowedBucketSchemes` slice is the source of truth and cannot be mutated through the accessor. `TestAllowedBucketSchemes_ReturnedSliceIsClone` pins it.
- [ ] **`Sources` is the Go field name; YAML key is `stateSources`.** Avoids the package-level naming collision with the `StateSources` struct itself while preserving the on-disk schema.
- [ ] **`validateNodeKeysMatch` enforces set equality between `index` and `paths`.** Sorted alphabetically before reporting, so the same node name appears in the error on every run. Two dedicated test cases pin both directions (missing from paths; missing from index).
- [ ] **Cross-source bucket uniqueness.** Duplicate buckets across `stateSources[]` are rejected with a message naming the previous index, mirroring the duplicate-destination pattern in #533.
- [ ] **Same node ID across multiple buckets is allowed.** That's the whole point of multi-bucket redundancy. `TestParseStateSources_SameNodeAcrossBucketsAllowed` pins it.
- [ ] **Empty `stateSources:` is valid.** Some networks may not use fast-sync; the parser must not require at least one source. `TestParseStateSources_EmptyStateSourcesTolerated` pins it.
- [ ] **A stray top-level `version:` field is caught by the schemaVersion validator**, not by strict-decode. `version:` is silently ignored at the schemaVersion stage (the validator inspects only `schemaVersion`), so the manifest surfaces as `MissingSchemaVersionError`. `TestParseStateSources_RejectsTopLevelVersionField` pins the contract.

## Test commands

```bash
# Unit tests (this PR brings the package to 97.8% coverage with the parser + 30 tests)
go test -race -cover -tags='!integration' ./pkg/manifests/...

# Lint pipeline (errorx gate, gofmt)
task lint:check
```

Expected:

```
ok      github.com/hashgraph/solo-weaver/pkg/manifests    coverage: 97.8% of statements
```

## Manual UAT

No CLI surface, no daemon hook. The fixture at the top of `pkg/manifests/state_sources_test.go` (`fullStateSourcesManifest`) is the canonical reference shape, mirroring the example in the HIP for `state-sources.yaml`. The intended user-facing behaviour surfaces once a follow-on story wires this parser into the daemon's fast-sync bootstrap flow.

#### Story 00535

# #535 — Review guide: schemaVersion validation across all manifests

> **Issue:** https://github.com/hashgraph/solo-weaver/issues/535
> **Epic:** https://github.com/hashgraph/solo-weaver/issues/501
> **Branch:** `00535-manifest-schema-version-validation`
> **Base:** `00501-manifest-parsing`

## Problem and solution

Epic #501 introduces parsers for the four YAML files under `manifests/` in the CN deployment package. Every one of those files declares a `schemaVersion`, and the epic spec requires that an unknown value is rejected with a clear error *before* any further parsing. Story #535 is the cross-cutting prerequisite: do this once in a shared package so the four parser stories (#531/#532/#533/#534) all plug into a single validator instead of each rolling their own.

This PR lays down a new `pkg/manifests/` package whose only consumer-facing entry point is the schemaVersion validator `ValidateSchemaVersion(kind, data)`; supporting types (`Kind`, `SchemaVersion`, `Header`, `SchemaV1`), error classifications, and the `SupportedVersions` helper round out the surface area but exist solely to serve that one validator. There is no production caller yet — concrete parsers come in #531–#534, each as its own PR targeting `00501-manifest-parsing`. The epic branch is merged to `main` only after all six stories land.

The validator inspects only the `schemaVersion` field (other fields are tolerated at this stage so a future parser can use strict-decode on the *typed* root struct without surprise). `schemaVersion` is an `int` per HIP-XXXX0 — the YAML carries `schemaVersion: 1`, not a string — and the validator probes with a pointer-typed scratch struct so it can distinguish "field absent" (nil → `MissingSchemaVersionError`) from "explicit value 0" (parsed but unsupported → `UnsupportedSchemaVersionError`). The set of accepted versions is a per-kind map seeded with `{1}` for all four kinds; widening to `2` later is a one-line change.

## Changed files

| File | What changed |
|---|---|
| `pkg/manifests/manifests.go` (new) | `Kind`, `SchemaVersion`, `Header`, `supportedVersions` registry, `ValidateSchemaVersion`, `SupportedVersions` |
| `pkg/manifests/errors.go` (new) | `errorx` namespace + four error types (`ParseError`, `MissingSchemaVersionError`, `UnsupportedSchemaVersionError`, `UnknownKindError`) and their constructors |
| `pkg/manifests/manifests_test.go` (new) | Table-driven unit tests; 96.7% coverage |
| `docs/claude/reviews/00535-manifest-schema-version-validation.md` (new) | This guide |

## Code review checklist

- [ ] **Errors are classifiable by `errorx.IsOfType`.** Every constructor returns a typed `*errorx.Error` so downstream code (and tests) can branch on classification, mirroring the pattern used by `pkg/software/errors.go` and `pkg/security/principal/errors.go`.
- [ ] **No strict-decode at this layer.** `TestValidateSchemaVersion_ToleratesUnknownFields` pins the contract: extra top-level fields do *not* trigger an error here. This is deliberate — per-kind parsers in #531–#534 will apply strict decoding against their typed root struct, where field-level errors are useful rather than misleading.
- [ ] **`schemaVersion` is decoded as an integer, not a string.** Aligns with HIP-XXXX0 and avoids any string→int round-trip downstream. A string-typed value (`schemaVersion: "v1"`) fails decoding with `ParseError`; `TestValidateSchemaVersion_NonIntegerValue` pins that contract.
- [ ] **Empty `schemaVersion:` is treated as missing.** YAML `schemaVersion:` (no value) decodes to a nil pointer in the probe struct and is rejected with `MissingSchemaVersionError`, not silently accepted as the zero value.
- [ ] **Unsupported value error message includes the supported set.** `unsupported_schema_version: manifest "external-files" declares schemaVersion 2 (supported: [1])` — both the offending value and the accepted list are in the message, and stored as printable error properties for structured logging.
- [ ] **`Kind` defensive check.** `UnknownKindError` covers the programmer-mistake path where a caller passes an unregistered `Kind`. Not a user-input error, but worth surfacing explicitly so future parser PRs catch it during development.
- [ ] **`sortedSupported` deterministic ordering.** Map iteration order is randomised in Go; the helper sorts before returning so error messages and `SupportedVersions(kind)` are reproducible. (Coverage on the comparator is 83.3% only because we ship a single version today — the comparator branch will be exercised once a second version lands.)

## Test commands

```bash
# Unit tests (the new package has no platform-specific code)
go test -race -cover -tags='!integration' ./pkg/manifests/...

# Lint pipeline (errorx gate, gofmt)
task lint:check
```

Expected:

```
ok      github.com/hashgraph/solo-weaver/pkg/manifests    coverage: 96.7% of statements
```

## Manual UAT

This PR introduces no CLI surface and no production caller, so there is no end-user UAT. The validator is exercised exclusively by its unit tests until a parser story (first one being #531) wires it in. Reviewers can inspect the API shape by reading `pkg/manifests/manifests.go` top-to-bottom — the package comment names the four manifest kinds and the deferred-to-follow-on stories.

#### Story 00536

# #536 — Review guide: destination marker-prefix enforcement on external-files.yaml

> **Issue:** https://github.com/hashgraph/solo-weaver/issues/536
> **Epic:** https://github.com/hashgraph/solo-weaver/issues/501
> **Branch:** `00536-destination-marker-prefix`
> **Base:** `00534-state-sources-parser` (stacked; auto-retargets up the chain as upstream PRs merge)

## Problem and solution

In #639 (story #533), the `external-files.yaml` parser validated that the `destination` field was non-empty but accepted any string. That left a hole: a council-signed manifest could declare a destination like `/etc/cron.d/x.sh` and the downloader would faithfully drop a file there. The scope-boundary test `TestParseExternalFiles_DestinationPrefixNotYetEnforced` pinned that deferral explicitly.

This PR closes the hole. The `destination` field must now start with one of the recognised marker prefixes (`HAPIAPP_DIR` or `SOLO_PROVISIONER_DIR`) followed by `/`. Arbitrary filesystem paths, unknown markers, and near-misses (e.g. `HAPIAPP_DIR_FOO/x`) all fail parsing with a clear error naming both the offending destination and the allowed set.

The scope-boundary test from #533 is **replaced** (not deleted) by a richer `TestParseExternalFiles_DestinationPrefixEnforcement` that pins both directions: positive cases for each allowed marker (including the `marker/` empty-subpath form, which is deliberately accepted because "sensible path under HAPIAPP_DIR" is the downloader's responsibility, not the parser's), and negative cases for absolute paths, relative paths, unknown markers, and near-misses.

### Stacked-PR layout

Top of the six-PR stack: `00501-…` ← #634 ← #636 ← #638 ← #639 ← #640 ← **this PR**. All six target the epic branch via the chain; GitHub auto-retargets each downstream PR as upstreams merge. This is the **final story** for epic #501 — once this lands and the chain unrolls into the epic branch, the epic merges to main.

## Changed files

| File | What changed |
|---|---|
| `pkg/manifests/external_files.go` | Added unexported `allowedDestinationPrefixes` slice (enforcement source of truth), exported `AllowedDestinationPrefixes()` accessor returning a cloned slice, and `validateDestinationPrefix(field, dest)` helper; wired into `ExternalFile.validate` after the non-empty check. New imports: `slices`, `strings` |
| `pkg/manifests/external_files_test.go` | Removed `TestParseExternalFiles_DestinationPrefixNotYetEnforced`; added `TestParseExternalFiles_DestinationPrefixEnforcement` (3 accept + 5 reject subtests), `TestAllowedDestinationPrefixes_PinnedSet`, and `TestAllowedDestinationPrefixes_ReturnedSliceIsClone` |
| `docs/claude/reviews/00536-destination-marker-prefix.md` (new) | This guide |

## Code review checklist

- [ ] **`AllowedDestinationPrefixes()` is exported and returns a clone.** External tooling (e.g. CN deployment-package linters) can call it to reuse the closed set without forking the list. The exported accessor wraps an unexported `allowedDestinationPrefixes` slice (the source of truth read by `validateDestinationPrefix`) so callers cannot weaken enforcement by mutating the returned slice — `TestAllowedDestinationPrefixes_ReturnedSliceIsClone` pins that contract.
- [ ] **Trailing `/` is required.** Without it, `validateDestinationPrefix("HAPIAPP_DIR")` would pass — bad. With it, the marker must be followed by `/`. `TestParseExternalFiles_DestinationPrefixEnforcement` pins both the rejection of `HAPIAPP_DIR` (alone) and `HAPIAPP_DIR_FOO/x` (prefix-of-longer-token).
- [ ] **`HAPIAPP_DIR/` is accepted (empty subpath).** Marker validation stops at "is the marker recognised". Whether the resulting path is sensible under that marker is the downloader's call, not the parser's. The test names this contract explicitly.
- [ ] **Error message lists the allowed set.** When a destination is rejected, the operator sees both the offending value and the prefixes they could've used — saves a round-trip to the docs.
- [ ] **The scope-boundary test from #533 is gone.** `TestParseExternalFiles_DestinationPrefixNotYetEnforced` was the explicit pin saying "this enforcement is deferred to #536." It's now replaced by the positive enforcement tests — the deferral is satisfied, the boundary is closed.
- [ ] **`TestAllowedDestinationPrefixes_PinnedSet` is the API surface guardrail.** Widening or narrowing the closed set requires deliberately changing that test — surfaces as a focused diff during review.

## Test commands

```bash
# Unit tests (97.9% coverage)
go test -race -cover -tags='!integration' ./pkg/manifests/...

# Targeted: just the new enforcement tests
go test -race -tags='!integration' -run 'DestinationPrefix|AllowedDestinationPrefixes' ./pkg/manifests/...

# Lint pipeline
task lint:check
```

Expected:

```
ok      github.com/hashgraph/solo-weaver/pkg/manifests    coverage: 97.9% of statements
```

## Manual UAT

No CLI surface; the only behavioural change is that an `external-files.yaml` shipping a destination outside the closed marker set will now fail parsing. Worth sanity-checking against any in-flight package fixtures that callers might be testing with — none currently exist in the repo (the parsers have no production caller yet, per the deliberate scope of epic #501).

### Related Issues

* Closes #501
* Closes #531
* Closes #532
* Closes #533
* Closes #534
* Closes #535
* Closes #536

## Test plan

- [x] `go test -race -cover -tags='!integration' ./pkg/manifests/...` — all green, 97.9% statement coverage
- [x] `task lint:check` — clean
- [ ] CI passes on this PR
- [ ] Reviewer reads `pkg/manifests/` top-down against [the epic demo doc](docs/claude/reviews/00501-manifest-parsing.md), then drills into individual story review docs as needed.


pkg/manifests/manifests.go
```diff
@@ -0,0 +1,103 @@
...
+	"sort"
+
+	"gopkg.in/yaml.v3"
+)
+
+// SchemaVersion is the value of the schemaVersion field on a manifest.
+type SchemaVersion string
+
+// SchemaV1 is the only schemaVersion currently accepted on any manifest.
+const SchemaV1 SchemaVersion = "v1"
```
leninmehedy: I believe DevOps wanted `SchemaVersion` to be an integer to avoid string -> int conversion or parsing. I am not sure what would be the best, but in the HIPs I converted it to int.

Let's discuss before we may any change.

pkg/manifests/state_sources.go
```diff
@@ -0,0 +1,169 @@
...
+
+// BucketType enumerates the cloud-storage backends recognised today.
+// Extending this set requires adding the case here and (separately) wiring
+// the downloader to authenticate against the new backend.
+type BucketType string
+
+const (
+	BucketTypeGCS BucketType = "gcs"
+	BucketTypeS3  BucketType = "s3"
+)
```
leninmehedy: seems you have worked off an old version of HIP. Can you please sync and update. I think the latest one doesn't have the Type filed anymore. However, if you feel having a type file makes the code simpler, we can update the HIP

leninmehedy (changes requested):
Let's discuss about my comments. It appears you probably haven't synced with the latest HIPs. The older version were fine, but then I tried to adapt based on comments from DevOps team. We however can go back to our original design if we can justify our original version is better.

pkg/manifests/manifests.go
brunodam: Agreed — switched `SchemaVersion` to `int` in 77b1c6c to match the HIP (`schemaVersion: 1`). `ValidateSchemaVersion` now probes with a pointer-typed scratch struct so we keep the missing-vs-unsupported error split: nil → `MissingSchemaVersionError`, parsed-but-unsupported (including explicit 0) → `UnsupportedSchemaVersionError`. A string value like `"v1"` now fails with `ParseError` and is pinned by `TestValidateSchemaVersion_NonIntegerValue`.

pkg/manifests/state_sources.go
brunodam: Synced — dropped the `type` field in 77b1c6c. Provider is now encoded in the bucket URI scheme (`gcs://`, `s3://`) per the HIP §"Bucket URIs use native cloud storage schemes". `validateBucketScheme` enforces the closed set with the same pattern as the destination-marker allowlist in #536; `AllowedBucketSchemes()` is exported (cloned slice) so package linters can reuse the policy. New test cases pin the unsupported-scheme, HTTPS-style, and scheme-alone failure modes.

leninmehedy (approved):
Thank you.

Issue #669, opened 2026-06-11
Cilium loadBalancer.acceleration: best-effort attaches XDP to the host public NIC, flapping carrier on ixgbe (X550) and dropping the node's public IP


nathanklick:
## Summary

`kube cluster install` installs Cilium with `loadBalancer.acceleration: "best-effort"` and `routingMode: native` **without an explicit `devices:` list**. On hardware whose primary/public NIC is an Intel X550 (`ixgbe`), Cilium auto-detects that public, default-route NIC as a datapath device and attaches an **XDP** program to it. The `ixgbe` driver performs a PHY/link reset whenever an XDP program is attached or the agent reconciles, which **flaps the NIC's carrier continuously** (Gained → Lost every ~10s). Because the link never holds carrier long enough for `systemd-networkd` to finish configuring it, the node's **static public IP is never installed and the host loses public connectivity**.

There is currently **no supported way to override the Cilium values** (acceleration, devices, etc.) through `kube cluster install`, because the values file is embedded in the binary and rewritten on every run (details below).

## Environment

- solo-weaver: `kube cluster install` flow (`internal/workflows/steps/step_cilium.go`)
- Cilium: **1.18.1** (`installCiliumCNI("1.18.1")`), installed via `cilium install --wait --version "1.18.1" --values <sandbox>/etc/weaver/cilium-config.yaml`
- Node type: `block` (single-node kubeadm cluster), but this affects any node-type on the same hardware
- NIC: Intel Ethernet Controller **X550**, driver **`ixgbe`**, on the host's public/default-route interface (`eno1`)
- OS: Ubuntu, `systemd-networkd` renderer, netplan with a static public `/31` on the public NIC

## Root cause

The embedded values template (`internal/templates/files/cilium/cilium-config.yaml`) sets:

```yaml
# Routing Configuration
routingMode: native
autoDirectNodeRoutes: true
# (no `devices:` key — Cilium auto-detects datapath devices)

# Load Balancer Configuration
loadBalancer:
  mode: dsr
  dsrDispatch: opt
  algorithm: maglev
  acceleration: "best-effort"   # <-- attaches XDP to native devices where the driver supports it
```

With `acceleration: best-effort` and no `devices:` scoping, Cilium attaches an XDP program to every auto-detected native device — including the host's public, default-route NIC. `ixgbe` resets the link on XDP attach/reconfigure, so the public NIC's carrier flaps indefinitely.

## Evidence

[...]

PR #674, opened 2026-06-11
fix(cilium): disable XDP load-balancer acceleration to prevent NIC carrier flaps

nathanklick:
## Problem

`kube cluster install` renders the embedded Cilium values with
`loadBalancer.acceleration: "best-effort"` and `routingMode: native` (no explicit
`devices:`). With `best-effort`, Cilium auto-detects native datapath devices —
including the host's **public/default-route NIC** — and attaches an **XDP**
program to each.

On `ixgbe` NICs (Intel X550) the driver resets the PHY whenever an XDP program is
attached or the agent reconciles. On a block-node host this flapped the public
NIC's carrier continuously (Gained → Lost every ~10s), with two consequences:

1. `systemd-networkd` never finished configuring the link, so the node's static
   public IP was never installed — the host lost public connectivity.
2. Sustained flapping tripped the **upstream switch's link-flap err-disable**,
   which shut the port. That state lives on the switch and **survived a host
   reboot**, so the host stayed offline until the provider re-enabled the port.

Because the reconcile daemon re-applies the embedded config, there is no durable
operator workaround — killing/reconfiguring Cilium by hand is reverted. The fix
has to be in the shipped template.

Full root-cause write-up and evidence: #669.

## Fix

Set `loadBalancer.acceleration: "disabled"` so Cilium uses **tc-eBPF** instead of
XDP. tc-eBPF performs the same DSR / maglev load balancing without attaching an
XDP program to any NIC, so the `ixgbe` PHY is never reset and the public NIC
cannot flap.

## Trade-off

Removes the XDP hardware-offload fast path (a throughput optimization). Preserves
full kube-proxy-replacement / NodePort / LoadBalancer behavior. Safe default
across mixed hardware.

## Follow-ups (not in this PR)

- Expose/scope an explicit `devices:` list so the datapath never binds the public
  NIC at all.
- The template sets `k8sServiceHost: {{.MachineIP}}`, and `GetMachineIP()` returns
  the first UP IPv4 (often the public NIC); device-scoping interacts with this, so
  a node-IP knob may be warranted.

## Test

```bash
grep -A1 acceleration <sandbox>/etc/weaver/cilium-config.yaml      # acceleration: "disabled"
KUBECONFIG=/etc/kubernetes/admin.conf kubectl -n kube-system \
  get cm cilium-config -o jsonpath='{.data.bpf-lb-acceleration}'   # disabled
# no XDP attached to any host NIC; public NIC keeps carrier (no Gained/Lost loop)
```

See `docs/claude/reviews/00669-disable-xdp-acceleration.md` for the full review guide and UAT.

### DIFF
--- a/docs/claude/reviews/00669-disable-xdp-acceleration.md
+++ b/docs/claude/reviews/00669-disable-xdp-acceleration.md
@@ -1,0 +1,94 @@
+# 00669 — Disable Cilium XDP load-balancer acceleration
+
+## Problem
+
+`kube cluster install` rendered the embedded Cilium values
+(`internal/templates/files/cilium/cilium-config.yaml`) with
+`loadBalancer.acceleration: "best-effort"` and `routingMode: native` with no
+explicit `devices:` list. With `best-effort`, Cilium auto-detects native
+datapath devices — including the host's public/default-route NIC — and attaches
+an **XDP** program to each.
+
+On `ixgbe` NICs (e.g. Intel X550) the driver performs a PHY/link reset whenever
+an XDP program is attached or the agent reconciles. This flapped the public NIC's
+carrier continuously (Gained → Lost every ~10s). Two consequences observed on a
+block-node host:
+
+1. `systemd-networkd` never finished configuring the link, so the node's static
+   public IP was never installed — the host lost public connectivity.
+2. Sustained flapping tripped the **upstream switch's link-flap err-disable**,
+   which shut the port. That state lives on the switch and **survives a host
+   reboot**, so the host stayed offline until the provider re-enabled the port.
+
+Because the reconcile daemon re-applies the embedded config, there was no durable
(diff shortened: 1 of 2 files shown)


brunodam:
Great fix! One thing worth noting for existing deployments: the Cilium step has an idempotency guard (`IsConfigured` persisted to disk + `cilium status` check), so nodes that already have a running cluster won't pick up the new template automatically — they'll keep `best-effort` until manually updated.

The good news is the migration framework handles this cleanly. A follow-up PR could add a `CLIVersionMigration` (scope: startup, gated on the version boundary where this ships) that re-renders `cilium-config.yaml` from the updated template and runs `cilium upgrade --values <config-path>`. Startup migrations run on every CLI invocation, so it would apply automatically the next time the provisioner runs on an existing node.

Happy to open a follow-up issue to track that if useful!

brunodam (dismissed):

PR #688, opened 2026-06-12
feat: solo provisioner daemon core

### LINKED ISSUE #499: Epic A38_2 — solo-provisioner-daemon Core
Implement the long-running daemon process that runs as a systemd service on the host.

leninmehedy:
## Description

A long-running, self-healing control plane that sits on every node, watches the cluster for operator intent, and executes upgrades and migration soaks autonomously — without ever taking itself down.

## Features
- **It refuses to fail quietly.** Bad config? It won't start, and systemd tells you instantly. A monitor crashes? It's caught, backed off, and restarted — the process never dies from a single subsystem fault. You always have a daemon to ask "what's wrong?"
- **It does each thing exactly once.** Upgrades are deduplicated three ways, keyed on the cluster's own source of truth. Restart it, re-deliver the event, replay the watch — it won't double-execute. Migration soaks are single-flight and resume after a reboot with elapsed time preserved, because progress is on disk, not in memory.
- **It tells you how to fix it, not just that it broke**. Every error in /status carries a machine reason, a human message, a copy-pasteable sudo … resolution, and a "since" timestamp. No journal spelunking for the common cases.
- **Zero new attack surface**. Local Unix socket only — no open port. Each component gets its own scoped Kubernetes credentials, so a block-node bug can't touch consensus-node resources. The privileged binary is deliberately tiny: it shares nothing with the CLI.
- **It plugs into the monitoring you already run.** Structured reason-tagged logs flow to journald → Grafana Loki through Alloy with zero daemon-side config; metrics go to Prometheus the same way. Every milestone is also written to a tamper-evident, fsync'd JSONL audit trail for postmortems.
- **It's built to grow.** Adding a new component or endpoint is a documented recipe that never touches existing routes. Config format changes are forward-safe via sealed, single-step migrations — an old daemon will never misread a new file.

Please review the daemon architecture guide: docs/daemon/daemon-architecture.md

Screenshot 2026-06-12 at 9 20 55 pm

### Related Issues

* Closes #499


brunodam (approved):
Amazing work!

PR #690, opened 2026-06-12
fix(cilium): restart agents via a separate v0.19.2 migration

nathanklick:
## Problem

Follow-up to #689 (shipped in v0.19.1). That migration runs `cilium upgrade --values` to flip existing clusters from `loadBalancer.acceleration: best-effort` → `disabled` (#669/#674). Verified on the testnet `blk` hosts after rolling v0.19.1:

- The migration ran and the `cilium-config` ConfigMap flipped to `disabled` (helm release `cilium` → rev 2) ✅
- **But XDP stayed attached to the public NIC** (`prog/xdp … cil_xdp_entry` on `eno1`) and the cilium agent never restarted (age 25h, 0 restarts) ❌

Root cause: the Cilium chart's agent pod template has **no `checksum/config` annotation**, so a ConfigMap-only `helm upgrade` doesn't roll the DaemonSet. The agent reads `bpf-lb-acceleration` only at startup — staged but never applied.

## Fix — a *separate* migration tied to v0.19.2

The restart is its **own** migration (`CiliumAgentRestartMigration`, `minVersion = 0.19.2`), not folded into the v0.19.1 migration — because the `blk` hosts already ran the v0.19.1 migration (`installed == 0.19.1`), so it won't re-fire on them. A 0.19.2-gated migration fires on the **0.19.1 → 0.19.2** upgrade and restarts the staged-but-not-applied agents.

- Registered **after** `CiliumAccelerationMigration`, so a `0.18.x → 0.19.2` upgrade flips the config then restarts in one pass.
- `Execute` restarts only when k8s + cilium are installed **and** acceleration is already `disabled` (the acceleration migration owns flipping the config); no-op otherwise.
- New `kube.Client.RolloutRestart(ctx, KindDaemonSet, "kube-system", "cilium")` (via the dynamic client; adds `KindDaemonSet` + GVR) + a `WaitForResource`/`daemonSetRolledOut` rollout wait that tolerates a transient NIC blip during XDP detach.
- A failed restart fails `Execute`. The v0.19.1 acceleration migration is unchanged.

Review guide: `docs/claude/reviews/00669-cilium-migration-agent-restart.md`.

### DIFF
--- a/cmd/cli/commands/root.go
+++ b/cmd/cli/commands/root.go
@@ -124,6 +124,7 @@ func RegisterMigrations() {
 	migration.Register(migration.ScopeStartup, state.NewHelmReleaseSchemaV2Migration())
 	migration.Register(migration.ScopeStartup, workflows.NewLegacyBinaryMigration())
 	migration.Register(migration.ScopeStartup, workflows.NewCiliumAccelerationMigration())
+	migration.Register(migration.ScopeStartup, workflows.NewCiliumAgentRestartMigration())
 
 	// ── Block-node upgrade migrations (run during block node upgrade workflow) ─
 	migration.Register(migration.ScopeBlockNode, blocknode.NewVerificationStorageMigration())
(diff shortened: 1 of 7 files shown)


brunodam (approved):
LGTM

PR #776, opened 2026-06-25
fix: vm lifecycle on UTM and auto-reload UTM

radkomih:
## Description

Encountered this multiple times: 

Task vm:start (and vm:reset, which calls it) intermittently hangs or falsely reports a stopped VM as running. Three root causes, all stemming from treating utmctl exit codes as a proxy for UTM state:
- Exit codes don't reflect VM state. utmctl status exits 0 even when the VM is stopped (only a missing VM is non-zero), and utmctl start exits 0 even when it never launched the VM. The old vm:start gated on utmctl status … && echo "running", so a stopped VM looked "running."
- vm:reset raced UTM. It ran delete → clone → start separated by fixed sleeps, issuing the clone/start before UTM had finished updating its registry. Starting immediately after a clone can wedge UTM's Apple Event handler.
 - vm:download:golden left UTM stale. It copied the .utm bundle into UTM's Documents dir while UTM was running, which doesn't live-reload bundles - then aborted with a "please restart UTM manually" message, breaking any automated flow.

### Changes

All in taskfiles/vm.yaml:

- vm:start - gate on the reported run state (started / stopped) instead of exit codes. If the VM is still stopped, re-issue utmctl start (transient glitch), and only as a last resort relaunch UTM.app to clear a wedged start path before aborting with actionable guidance.
- vm:reset - after delete, poll until utmctl status reports the VM is gone; after clone, poll until it reports stopped - so the start doesn't race a half-registered VM.
  - vm:download:golden - add a reload_utm helper that quits/relaunches UTM and waits until it's scriptable again, then verifies UTM sees the golden VM and continues, instead of aborting and asking the user to
  restart UTM by hand.

### Testing

Verified manually

### DIFF
--- a/taskfiles/vm.yaml
+++ b/taskfiles/vm.yaml
@@ -126,15 +126,19 @@ tasks:
           pkill -f "UTM.app/Contents/MacOS/UTM" >/dev/null 2>&1 || true
           sleep 2
           open -g -a UTM || true
-          # Wait until UTM is scriptable again before relying on utmctl.
+          # Return non-zero if UTM never becomes scriptable, so callers can tell
+          # "UTM never came back" from "UTM is up but doesn't see the VM".
           for _ in $(seq 1 10); do
             if osascript -e 'with timeout of 5 seconds' \
                  -e 'tell application "UTM" to get name of every virtual machine' \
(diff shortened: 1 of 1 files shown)


brunodam (approved):
Thank you

PR #785, opened 2026-06-26
refactor(version): replace in-repo pkg/version with automa-saga/version

alex-au:
## Description

solo-weaver maintained a bespoke version package (`pkg/version` plus `cli`/`daemon`
subpackages) that used a `SetCurrent` registry and per-OS `go generate` scripts to write
`VERSION`/`COMMIT` files which were then `//go:embed`-ed into each binary. This PR replaces
that machinery with the shared, dependency-free library
[`github.com/automa-saga/version@v1.0.0`](https://github.com/automa-saga/version), whose
`Info` is stamped at build time via `-ldflags -X` — no embed files, no generate scripts, no
registry.

Each binary keeps a tiny inline `version` Cobra command (the library is framework-agnostic),
and all internal callers read build metadata directly from the shared package. Observable CLI
output is unchanged except two intended deltas: the Go-version key is now `goVersion` (was
`goversion`) and a new `date` field is included. The daemon's `--version` JSON is still a single
object whose `version`/`commit` keys the daemon-install probe parses unchanged.

### Files changed

| File | Change |
|---|---|
| `cmd/cli/commands/version.go`, `cmd/daemon/version.go` | **New.** Inline `version` command + `renderVersion` (json default, yaml via `yaml.Marshal`, unsupported → `errorx.IllegalFormat`). |
| `cmd/cli/commands/version_test.go`, `cmd/daemon/version_test.go` | **New.** `renderVersion` table tests: json (default/explicit/case), yaml, and the unsupported-format error. |
| `cmd/cli/commands/root.go` | `--version` flag + `version` subcommand use the inline command; dropped `pkg/version/cli` import. |
| `cmd/daemon/main.go` | Import → `automa-saga/version`; inline `version` cmd; pre-cobra `--version` path uses `version.Get().Format(FormatJSON)`. |
| `cmd/cli/commands/common/run.go`, `internal/state/state_manager.go`, `internal/ui/logging.go`, `internal/workflows/steps/step_weaver.go` | `version.Number()/Commit()` → `version.Get().Version/.Commit`. |
| `internal/state/state.go`, `internal/ui/view.go` | `Info.Number` → `Info.Version`. |
| `internal/doctor/diagnose.go` | `version.Commit()/Number()` → single `version.Get()` snapshot. |
[...]

PR #795, opened 2026-07-01
fix(software): retry truncated/empty downloads and report real checksum algorithm

### LINKED ISSUE #786: Software downloads should retry on truncated/empty files instead of failing the run
## Problem

Software downloads (`pkg/software/downloader.go`) fail the whole workflow when a download
comes back truncated or empty, with no retry. The checksum verifier *correctly* rejects the
bad file, but because there is no retry the entire `kube cluster install` (and any UAT built on
it) aborts on a transient CDN/network hiccup.

Observed in the UAT Purge Storage job on PR #785
([run](https://github.com/hashgraph/solo-weaver/actions/runs/28215368958/job/83586822330)),
during `setup-crio`:

```
software.checksum_error: checksum verification failed for file
  '/opt/solo/weaver/downloads/cri-o.amd64.v1.33.4.tar.gz' using algorithm 'unknown'
  [ expected = 8f6d14828659b85da7c83bad798d50c2f7e0311742615fb7ed305f77bab54e50
    actual   = e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 ]
task: Failed to run task "uat:setup": exit status 1
```

The `actual` value `e3b0c442…b855` is the SHA-256 of a **zero-byte file**, i.e. the cri-o
tarball was written empty. `Downloader.Download` (downloader.go:135) does a single `http.Get`,
checks only `StatusCode == 200`, then `io.Copy`s the body to disk — so a 200-with-empty-body, a
silently truncated copy, or a mid-stream reset produces a corrupt file that the run cannot
recover from.

## Proposed fix

Make downloads self-healing for transient corruption:

- **Retry with backoff** around `Download` (the repo already vendors `cenkalti/backoff`). Treat
  download errors *and* a subsequent checksum mismatch as retryable: on mismatch, delete the bad
  file and re-fetch up to N attempts before giving up.
- **Validate the download before trusting it**: reject zero-byte files, and when the response
  carries `Content-Length`, fail fast (and retry) if bytes written ≠ advertised length, rather
  than persisting a truncated file.
- **Minor cleanup**: `integrity_checker.go` hard-codes `algorithm: "unknown"` in the mismatch
  error (lines ~39/44) even though sha256 was used — thread the real algorithm through so the
  error message isn't misleading.

## Acceptance

- [ ] A transient empty/truncated download is retried automatically and succeeds without failing the workflow.
- [ ] A persistently-corrupt download (checksum never matches) still fails after exhausting retries, with a clear error.
- [ ] Zero-byte / `Content-Length`-mismatched responses are detected and retried, not persisted.
- [ ] Checksum-mismatch errors report the actual algorithm used (no more `algorithm 'unknown'` for sha256).

alex-au:
## Description

Software downloads (`pkg/software/downloader.go`) failed the whole workflow when a
download came back truncated or empty, with no retry. `Download` did a single `http.Get`,
checked only `StatusCode == 200`, then `io.Copy`'d the body to disk — so a 200-with-empty-body,
a silently truncated copy, or a mid-stream reset produced a corrupt file that the checksum
verifier correctly rejected but the run could not recover from. Observed in the UAT Purge
Storage job on PR #785 during `setup-crio`:

```
software.checksum_error: checksum verification failed for file
  '/opt/solo/weaver/downloads/cri-o.amd64.v1.33.4.tar.gz' using algorithm 'unknown'
  [ expected = 8f6d1482...b54e50
    actual   = e3b0c442...b855 ]   # sha256 of a zero-byte file
```

Downloads are now self-healing for transient corruption. `Download` validates the response
before trusting it (rejects zero-byte bodies, and — when `Content-Length` is advertised — fails
on a byte-count mismatch instead of persisting a truncated file). A new
`Downloader.DownloadAndVerify` wraps download + checksum in a bounded retry loop: both download
errors and a subsequent checksum mismatch are treated as retryable, the bad file is deleted
between attempts, and it gives up after `maxAttempts` (default 3) with capped exponential
backoff. The three `base_installer.go` call sites now use it.

The misleading `algorithm 'unknown'` in checksum-mismatch errors is also fixed — the real
algorithm (`sha256`, etc.) is now threaded through the checksum helper.

Retry uses a small hand-rolled loop matching the repo's existing `ProbeTCP` style rather than
pulling in `cenkalti/backoff` (which is only a transitive go.sum entry, not a vendored/used
dependency).

### Files changed

| File | Change |
|------|--------|
| `pkg/software/downloader.go` | `Download` rejects zero-byte / Content-Length-mismatched responses; new `DownloadAndVerify` retry loop + `WithMaxAttempts`/`WithRetryDelay` options |
| `pkg/software/integrity_checker.go` | Thread the real algorithm into the `checksum` helper so mismatch errors no longer report `unknown` |
| `pkg/software/base_installer.go` | Replace the three `Download`+`VerifyChecksum` pairs (archive/binary/config) with `DownloadAndVerify` |
| `pkg/software/downloader_test.go` | Tests: zero-byte + Content-Length-mismatch rejection, retry-then-succeed, retry-exhaustion |
| `pkg/software/integrity_checker_test.go` | Assert mismatch error names the real algorithm, never `unknown` |

### Review guide

**Code review checklist**
- [ ] `downloader.go` `Download` — zero-byte check fires before the Content-Length check; the Content-Length comparison is guarded on `resp.ContentLength >= 0` so chunked/unknown-length responses (`-1`) are not falsely rejected.
- [ ] `downloader.go` `DownloadAndVerify` — the bad file is removed on every failed attempt (via base-path-validated `os.Remove`), so a corrupt partial never persists; the last error is returned once attempts are exhausted.
- [ ] `downloader.go` `sleepBackoff` — delay is capped at `maxRetryDelay` and the shift-overflow guard (`delay < fd.retryDelay`) holds; a zero `retryDelay` disables sleeping (used by tests).
- [ ] Truncation/empty errors reuse `NewDownloadError` wrapping an `errorx.ExternalError` cause — they stay `software.download_error` (retryable), no new errorx namespace.
- [ ] `integrity_checker.go` — all three `checksum(...)` calls pass the matching algorithm string; both `NewChecksumError` sites use it.
- [ ] `base_installer.go` — the pre-existing-file skip/delete logic is untouched; only the download+verify pair was collapsed.

**Unit tests**
```bash
# macOS-safe (pkg/software has no Linux-only deps):
go test -race -tags='!integration' ./pkg/software/... \
  -run 'Test_Downloader_Download_ZeroByte|Test_Downloader_Download_ContentLengthMismatch|Test_Downloader_DownloadAndVerify|Test_IntegrityChecker_VerifyChecksum'

# Or the full suite (run in the UTM VM for complete coverage):
task vm:test:unit
```

**Manual UAT**

The failure is inherently transient (CDN/network hiccup), so it is exercised by the
`Test_Downloader_DownloadAndVerify_RetriesThenSucceeds` test, which stands up an `httptest`
server that returns an empty body on the first request and the correct payload on the second,
and asserts the download recovers without failing. `..._ExhaustsAttempts` asserts a
persistently-corrupt file still fails after 3 tries with the file removed.

**Risks / rollback**

A persistently-unreachable or invalid URL now takes up to 3 attempts (plus a few seconds of
backoff) before failing, instead of failing immediately — bounded, and only on the error path.
The change is additive to the download path; revert is dropping `DownloadAndVerify` and
restoring the original `Download`+`VerifyChecksum` call-site blocks.

### Related Issues

* Closes #786

### DIFF
--- a/pkg/software/base_installer.go
+++ b/pkg/software/base_installer.go
@@ -201,14 +201,8 @@ func (b *baseInstaller) downloadAndVerifyArchive(archive ArchiveDetail) error {
 		}
 	}
 
-	// Download the archive
-	err = b.downloader.Download(downloadURL, destinationFile)
-	if err != nil {
-		return err
-	}
-
-	// Verify the downloaded archive's checksum
-	err = VerifyChecksum(destinationFile, checksum.Value, checksum.Algorithm)
+	// Download the archive and verify its checksum, retrying transient corruption
+	err = b.downloader.DownloadAndVerify(downloadURL, destinationFile, checksum.Value, checksum.Algorithm)
 	if err != nil {
 		return err
 	}
@@ -285,14 +279,8 @@ func (b *baseInstaller) downloadAndVerifyBinary(binary BinaryDetail) error {
 		}
 	}
 
-	// Download the binary
-	err = b.downloader.Download(downloadURL, destinationFile)
-	if err != nil {
-		return err
-	}
-
-	// Verify the downloaded binary's checksum
-	err = VerifyChecksum(destinationFile, checksum.Value, checksum.Algorithm)
+	// Download the binary and verify its checksum, retrying transient corruption
+	err = b.downloader.DownloadAndVerify(downloadURL, destinationFile, checksum.Value, checksum.Algorithm)
 	if err != nil {
 		return err
 	}
@@ -328,14 +316,8 @@ func (b *baseInstaller) downloadConfigs() error {
 			}
 		}
 
-		// Download the config file
-		err = b.downloader.Download(config.URL, configFile)
-		if err != nil {
-			return err
-		}
-
-		// Verify the downloaded config file's checksum
-		err = VerifyChecksum(configFile, config.Value, config.Algorithm)
+		// Download the config file and verify its checksum, retrying transient corruption
+		err = b.downloader.DownloadAndVerify(config.URL, configFile, config.Value, config.Algorithm)
 		if err != nil {
 			return err
 		}
--- a/pkg/software/downloader.go
+++ b/pkg/software/downloader.go
@@ -20,13 +20,22 @@ import (
 	"github.com/hashgraph/solo-weaver/pkg/sanity"
 )
 
+// Default retry settings for DownloadAndVerify.
+const (
+	defaultMaxAttempts = 3               // initial attempt + 2 retries
+	defaultRetryDelay  = 1 * time.Second // base delay; grows exponentially per attempt
+	maxRetryDelay      = 30 * time.Second
+)
+
 // Downloader is responsible for downloading a software package and checking its integrity.
 type Downloader struct {
 	client         *http.Client
 	timeout        time.Duration
-	basePath       string   // Base directory for validating download/extraction paths
-	allowedDomains []string // List of allowed domains for SSRF protection
-	insecureTLS    bool     // Skip TLS certificate verification (for local dev with self-signed certs)
+	basePath       string        // Base directory for validating download/extraction paths
+	allowedDomains []string      // List of allowed domains for SSRF protection
+	insecureTLS    bool          // Skip TLS certificate verification (for local dev with self-signed certs)
+	maxAttempts    int           // Number of download+verify attempts before giving up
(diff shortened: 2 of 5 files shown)


brunodam (dismissed):
This is great!

brunodam (approved):

PR #797, opened 2026-07-01
feat: verify binary checksums before execution

### LINKED ISSUE #340: Verify hash of binary before exec.RunCommand()

radkomih:
## Description

- Adds software.VerifyExecutables(artifactName) — re-hashes installed binaries on disk against their catalog checksums right before execution, catching post-install tampering that install-time checks miss (pkg/software/exec_verifier.go).
- Resolves checksums against the version recorded in on-disk state (via new state.ReadSoftwareVersionFromDisk), falling back to the catalog default, and skips with a warning when the installed version is no longer in the catalog.
- Adds a reusable VerifyExecutablesStep inserted before SetupSystemdService for kubelet, cri-o, and teleport.
- Adds inline verification before directly-exec'd binaries: cilium (install/rollback + acceleration migration), kubeadm/kubectl (InitializeCluster, ResetCluster best-effort), and teleport configure.
- Extracts shared crioInstalledBinaryPaths helper + CrioArtifactName constant so Install and verification never drift.

### Related Issues

- Closes #340


internal/workflows/migration_cilium_acceleration.go
```diff
@@ -117,6 +118,11 @@ func (m *CiliumAccelerationMigration) Execute(ctx context.Context, mctx *migrati
 		Str("to", disabledAcceleration).
 		Msg("Applying cilium-config (loadBalancer.acceleration=disabled) to existing cluster via 'cilium upgrade'")
 
+	// Verify the cilium binary before executing it.
+	if err := verifyCiliumExecutable(); err != nil {
```
alex-au, line 122: This is great, should we do the same [before L124 in migration_cilium_host_legacy_routing.go](https://github.com/hashgraph/solo-weaver/blob/00340-verify-binary-checksums-before-execution/internal/workflows/migration_cilium_host_legacy_routing.go#L124-L128)?

alex-au (commented):
Looks great, some nitpicks in the comment. Can I also confirm if the "obsolete CiliumHostLegacyRoutingMigration" has been removed as stated in the PR description?

internal/workflows/migration_cilium_acceleration.go
radkomih: I added the verification check (the PR summary was not up to date)

alex-au (approved):
Nice!

PR #792, opened 2026-06-30
feat(daemon): scaffold traffic_shaper monitor skeleton in solo-provisioner-daemon

alex-au:
## Description

Stands up the block-node `traffic_shaper` monitor as a real `daemonkit`-supervised
component inside `solo-provisioner-daemon` — the structural skeleton that the rest of
epic **TS_3 (#736)** hangs off. Until now the monitor was a no-op stub that logged once
and blocked on `ctx`. It now runs its two long-lived responsibilities concurrently and
reports its health over the daemon socket.

`Run(ctx)` launches the **pod-lifecycle watcher** and the **statusz poll loop** as two
goroutines, each wrapped in a per-responsibility retry loop with **5 s → 5 min**
exponential back-off. A fault in one responsibility is logged and retried in place and is
never propagated to `Run`, so a subsystem fault can't stop the other responsibility, trip
the top-level supervisor, or kill the daemon process. Monitor health (supervisor
`running` / `backoff` / `stopped`) is exposed via `GET /block_node/traffic_shaper/status`.

This is **skeleton only**: both responsibility bodies are stubs. The real veth resolution,
ingress HTB attach, statusz client, category→policy diff, nft membership apply, bootstrap
sequencing, and privileged-exec delegation land in sibling stories (#747–#755, #774). The
`degraded` health state and alerting are owned by #750 — this PR reports only supervisor
state.

### Files changed

| File | Change |
|------|--------|
| `internal/daemon/blocknode/traffic_shaper_monitor.go` | Export `TrafficShaperMonitor`; `Run` launches the two responsibilities via `superviseResponsibility` (5 s → 5 min back-off, faults swallowed); stub `runPodWatcher` / `runStatuszPoll` bodies |
| `internal/daemon/blocknode/handler.go` | New `BlockNodeHandler` (`daemonkit.ComponentHandler`) serving `GET /block_node/traffic_shaper/status`; 503 when monitor disabled |
| `internal/daemon/blocknode/component.go` | `ComponentResult` carries `*TrafficShaperMonitor` for handler wiring |
| `internal/daemon/daemon.go` | Block-node block builds the `trafficShaperStateFn` closure over the component `StatusTracker` and registers `BlockNodeHandler` |
| `internal/daemon/blocknode/traffic_shaper_monitor_test.go` | Monitor unit tests: returns on ctx cancel; faults retried without propagating; clean re-entry |
[...]

PR #800, opened 2026-07-02
docs(claude): require Resolution hints on every StepFailureReport

brunodam:
## Description

Adds a mandatory coding standard to `CLAUDE.md`: every `automa.StepFailureReport`
must include a `WithProperty(models.ErrPropertyResolution, []string{...})` block
listing operator-actionable steps to resolve the failure.

This pattern is already consistent across the codebase — this entry makes it
explicit so it is enforced on all future PRs rather than caught in review.

### Related Issues

* Introduced during #792 UAT review


alex-au (approved):
yes we should always do that

Issue #684, opened 2026-06-12
refactor(hardware): replace flat requirements lookup with per-node RequirementsProvider


brunodam:
## Goal

Replace the flat `(nodeType, profile) → BaselineRequirements` lookup table in `pkg/hardware/requirements.go` with a per-node-type `RequirementsProvider` interface driven by a structured `DeploymentSpec`. Each node type owns its own sizing rules; the engine composes contributions from independent rules instead of multiplying lookup-table axes.

Today the registry encodes requirements as a 2-D map keyed on `(nodeType, profile)`. This already mis-models reality:

- Block node `tier1-lfh` and `tier1-rfh` have the same 5 TB floor even though RFH with the `cloud-storage-archive` + `cloud-storage-expanded` plugins needs only ~150–200 GB of local disk.
- The mainnet / testnet / previewnet floors are inverted relative to documented intent (previewnet is currently the heaviest; mainnet matches testnet by accident).
- The only escape today is `--skip-hardware-checks`, which is the wrong primitive to lean on for production.

Adding axes to the table (preset, then features, then TPS/retention, then consensus-side knobs, then mirror-side knobs) leads to combinatorial blowup and duplicated cells. The fix is to stop modeling requirements as a lookup table and start modeling them as a **computation over a deployment spec** — one provider per node type, each free to use rules, tables, or formulas internally, behind a stable interface.

Once this is in place, the per-node install commands plumb their full configuration (`--profile`, `--plugin-preset`, `--plugins`, future feature flags) into a `DeploymentSpec`, and each provider's rules can decide the floor independently. This unblocks:

- Right-sizing Tier-1 RFH boxes (smaller local disk than LFH on the same profile).
- A correct mainnet floor that's not a duplicate of testnet.
- Future consensus-side or mirror-side feature flags that affect sizing without forcing a new column on every block-node row.

## Acceptance criteria

- `pkg/hardware` exposes `type RequirementsProvider interface { Compute(DeploymentSpec) (BaselineRequirements, error) }` and a `Providers()` registry keyed by node type.
- `DeploymentSpec` carries `Profile string`, `NodeType string`, and an open-ended `Options map[string]any` (or a typed equivalent) so callers can add new sizing inputs without changing the interface.
- A `blockNodeProvider` is registered for `models.NodeTypeBlock`. Its rules **reproduce today's numbers exactly** when `Options` is empty (regression-safe).
[...]

PR #807, opened 2026-07-07
refactor(cli): drop --profile and --node-type from kube cluster install

### LINKED ISSUE #685: refactor(cli): drop --profile and --node-type from kube cluster install
## Goal

Make `solo-provisioner kube cluster install` workload-agnostic by dropping the `--profile` and `--node-type` flags and replacing the workload-sized hardware preflight with a **substrate-only** floor — i.e. what Kubernetes itself needs to run on the host, independent of which Hedera node will eventually be deployed.

Today both flags exist only to drive `NewNodeSafetyCheckWorkflow(nodeType, profile, …)` at cluster-install time, which validates the host against the **eventual workload's** requirements (block-node testnet floor is 8 CPU / 16 GB / 5 TB). That's a defensive proxy: it tries to fail early on "is this host fit for the BN you'll later deploy?" at the cost of coupling cluster install to per-workload knowledge it shouldn't carry.

This coupling forces every new per-workload sizing input (plugin presets, feature flags, expected TPS / retention, future consensus or mirror knobs) to also pass through `kube cluster install`'s flag set — see the Tier-1 RFH right-sizing question in the recent #block-node-ops thread, where the desired RFH floor is ~200 GB but the cluster command demands 5 TB before any workload is even named.

Per-workload sizing belongs in the per-workload install commands (`block node install`, `block node check`, future `consensus node install`, etc.), where the full `DeploymentSpec` is known. `kube cluster install` should only verify "can this host run Kubernetes itself?"

Depends on #684 for a `"k8s-substrate"` provider that returns the substrate-only floor through the same `RequirementsProvider` interface used elsewhere.

## Acceptance criteria

- `solo-provisioner kube cluster install` no longer accepts `--profile` or `--node-type`. Both flags are removed from `cmd/cli/commands/kube/cluster/install.go` and any common-flag wiring local to that command path.
- Hardware preflight at cluster install validates **only** K8s substrate minimums — control-plane CPU/RAM, free disk under `/var/lib/`, supported OS, kernel module availability — via the `"k8s-substrate"` provider from #684.
- Per-node install / check commands (`block node check`, `block node install`, future `consensus node install`) continue to run the **full workload-sized preflight** against their per-node provider. They are the only entry points for workload-aware hardware validation.
- A host that meets K8s substrate minimums but not the previous workload-sized floor (e.g. a 100 GB-disk VM intended for a Tier-1 RFH BN) can now complete `kube cluster install --non-interactive` without `--skip-hardware-checks`. The mismatch surfaces later at `block node check`, with a clear workload-attributed message.
- `--skip-hardware-checks` semantics preserved: it continues to bypass whatever floor each command checks (substrate floor for cluster install; workload floor for node commands).
- `docs/quickstart.md` is updated: the `kube cluster install` example block, the **Additional Flags** table, and the surrounding "install K8s → install workload" narrative reflect the new shape. Per `CLAUDE.md` workflow rule 6, this is mandatory for CLI surface changes.
- `--help` output for `solo-provisioner kube cluster install` matches the new flag set and the doc.
- No silent change to per-node commands. Their existing per-profile floors continue to enforce until #684 lands the preset-aware sizing.

## Implementation notes

- `cmd/cli/commands/kube/cluster/install.go`:
  - Drop the `flagProfile` retrieval, the `IsValidProfile` guard, the `config.SetProfile(flagProfile)` call, and the `flagNodeType` resolution + `IsValidNodeType` guard.
  - Simplify the `workflows.InstallClusterWorkflow(...)` call to the new signature (no nodeType, no profile).
  - A scan for `config.GetProfile()` consumers in non-test code shows only `pkg/hardware/node_spec.go` — workload-side — so removing the `SetProfile` call from cluster install is safe. Confirm in the PR by re-running the grep against the final tree.
- `cmd/cli/commands/kube/cluster/cluster.go`: remove the `flagNodeType` variable and its flag registration.
- `internal/workflows/cluster.go`: change `InstallClusterWorkflow` signature to `(skipHardwareChecks bool, mr software.MachineRuntime)`. The downstream `kubernetesSetupWorkflow` already doesn't reference `nodeType` or `profile`.
- `internal/workflows/setup.go` / `internal/workflows/preflight.go`:
  - Either rename `NodeSetupWorkflow` → `SubstrateSetupWorkflow` for the cluster path and keep a separate node-typed setup for the per-node commands, **or** introduce a new `SubstrateSafetyCheckWorkflow` and have cluster install use that.
  - The new substrate workflow runs `CheckPrivilegesStep` + `CheckWeaverUserStep` + a single substrate-hardware step (OS + CPU + memory + disk against K8s control-plane numbers) — independent of node type and profile.
  - Per-node commands keep `NewNodeSafetyCheckWorkflow` and gain the workload-sized hardware steps from #684.
- The substrate floor numbers should live in the new `provider_substrate.go` from #684, not as a duplicated constant here.
- TUI / log strings: `node_setup` / `system_setup` IDs that include `nodeType` (`block-node-setup`, `block-system-setup`) should be renamed in the cluster-install path to something node-type-agnostic (e.g. `substrate-setup`, `substrate-system-setup`). Per-node commands' IDs are unaffected.

## Files to create / modify

| File | Change |
|------|--------|
| `cmd/cli/commands/kube/cluster/install.go` | Remove `--profile` / `--node-type` resolution and validation; simplify workflow call |
| `cmd/cli/commands/kube/cluster/cluster.go` | Drop `flagNodeType` and any local flag registration tied to it |
| `internal/workflows/cluster.go` | Update `InstallClusterWorkflow` signature; drop `nodeType` / `profile` params |
| `internal/workflows/setup.go` | Add substrate setup workflow; keep node-typed setup for node commands |
| `internal/workflows/preflight.go` | Add `SubstrateSafetyCheckWorkflow`; keep `NewNodeSafetyCheckWorkflow` for per-node commands |
| `internal/workflows/cluster_test.go` | Cover the new substrate path; remove tests that exercised the BN-sized preflight at cluster install |
| `docs/quickstart.md` | Update `kube cluster install` example, flag table, and narrative to reflect substrate-only validation |
| `internal/workflows/notify/...` (path TBD) | If phase/step labels reference "Node Setup" at cluster install, rename to "Substrate Setup" |

## Tests

- Unit: `kube cluster install` does not reject calls without `--profile` / `--node-type`; the flags are no longer registered.
- Unit: substrate preflight rejects hosts below K8s control-plane minimums; passes hosts that meet substrate-only but not the previous workload floor.
- Integration (`task vm:test:integration`): a small host that previously failed `kube cluster install --profile testnet --node-type block` on storage now completes the install; a follow-up `block node check --profile testnet --plugin-preset tier1-rfh` succeeds or fails per the workload floor — confirming the responsibility boundary moved correctly.
- Regression: per-node install / check commands still enforce per-profile (and post-#684, preset-aware) floors — no quiet relaxation of workload validation.
- Docs lint / link check (`task lint:check` or equivalent) passes on the updated `docs/quickstart.md`.

alex-au:
## Summary

Makes `kube cluster install` workload-agnostic. It now validates only the **Kubernetes substrate hardware floor** (2 CPU / 2 GB / 20 GB) via the `k8s-substrate` provider added in #684, instead of a block-node workload-sized preflight (16 CPU / 64 GB / 5 TB). The `--profile` and `--node-type` flags are **deprecated** — hidden from `--help` and ignored, but still accepted so existing scripts do not break.

Per-workload hardware validation stays where the full `DeploymentSpec` is known — `block node check` / `block node install` — so a small host intended for a Tier-1 RFH block node can stand up the cluster without `--skip-hardware-checks`; the workload mismatch surfaces later at `block node check`.

Closes #685. Depends on #684 (merged).

## Problem

`--profile` / `--node-type` existed on `kube cluster install` only to drive a workload-sized preflight at cluster-install time, coupling "stand up Kubernetes" to per-workload knowledge it should not own. Every new sizing axis (plugin presets, TPS/retention) would otherwise have to thread through the cluster command's flags.

## Design notes

- **`InstallClusterWorkflow(skipHardwareChecks, mr)`** is now substrate-only: `SubstrateSetupWorkflow` + `KubernetesSetupWorkflow`. Used solely by `kube cluster install`.
- **`block node install` no longer calls `InstallClusterWorkflow`.** Its bootstrap branch composes `NodeSetupWorkflow(block, profile, preset, skip)` + `KubernetesSetupWorkflow(mr)` directly, preserving the workload-sized preflight byte-for-byte. `kubernetesSetupWorkflow` was exported to enable this.
- **Substrate preflight reuses the node preflight's per-resource steps.** `NewSubstrateSafetyCheckWorkflow` runs privileges + weaver-user + the same four steps (`CheckOSStep`/`CheckCPUStep`/`CheckMemoryStep`/`CheckStorageStep`) with a substrate spec, and omits `CheckHostProfileStep`. Those steps route through `hardware.CreateNodeSpec`, which rejects node type `k8s-substrate` and an empty profile, so the preflight `createNodeSpec` wrapper sends the substrate type to `hardware.CreateSubstrateSpec` instead (resolves the floor via `NewNodeSpec` — registry-only, no IsValid gate). Each resource keeps its own TUI entry, failure report, and resolution hint (the shared steps previously carried only `WhyFloor`). Like the node preflight, it stops at the first failing hardware step — see UAT-3b.
- `SupportedNodeTypes()` / `SupportedProfiles()` are intentionally **not** polluted with substrate/empty entries.
- **Flag deprecation over hard removal.** `--profile`/`--node-type` are kept registered but hidden (`SetVarPHidden`, the same mechanism `alloy cluster` uses to retire flags) and their values are ignored; `install.go` warns via `cmd.Flags().Changed(...)` when either is explicitly passed. This keeps existing invocations working while `--help` shows the clean new surface.

## Changed files

| File | Change |
|------|--------|
| `pkg/hardware/provider_substrate.go` | Add `NodeTypeSubstrate` const; replace bare `"k8s-substrate"` literals |
| `pkg/hardware/factory.go` | New `CreateSubstrateSpec(host)` — bypasses node-type/profile IsValid gates |
| `internal/workflows/preflight.go` | New `NewSubstrateSafetyCheckWorkflow` reusing the four per-resource steps; `createNodeSpec` routes substrate to `CreateSubstrateSpec`; resolution hints added to the hardware steps |
| `internal/workflows/setup.go` | New `SubstrateSetupWorkflow` (substrate preflight + shared system setup) |
| `internal/workflows/cluster.go` | `InstallClusterWorkflow` substrate-only; export `KubernetesSetupWorkflow` |
| `internal/bll/blocknode/install_handler.go` | Bootstrap branch recomposed; workload preflight preserved |
| `cmd/cli/commands/kube/cluster/install.go` | Drop profile/node-type validation + `SetProfile`; simplify workflow call; register `--node-type` hidden + warn when deprecated flags are passed |
| `cmd/cli/commands/kube/cluster/cluster.go` | `flagNodeType` retained to back the hidden deprecated flag |
| `cmd/cli/commands/kube/kube.go` | `--profile` kept hidden/persistent (deprecated, ignored) instead of removed |
| `internal/workflows/cluster_it_test.go` | Update `InstallClusterWorkflow(false, nil)` call |
| `pkg/hardware/provider_substrate_test.go` | Test `CreateSubstrateSpec` bypass + floor |
| `internal/workflows/preflight_test.go` (new) | Assert substrate preflight composition (omits host-profile step) |
| `docs/quickstart.md`, `docs/dev/daemon/daemon-testing-guide.md`, `docs/dev/functionality-test-suite.md` | Docs for the new command shape |

## Review checklist

- [ ] `kube cluster install` runs substrate-only preflight; no `--profile`/`--node-type` on `--help`.
- [ ] Deprecated `--profile`/`--node-type` are still accepted (no `unknown flag`), ignored, and warn when passed.
- [ ] `block node install` bootstrap still runs the block-node workload preflight before Kubernetes setup, in that order.
- [ ] `CreateSubstrateSpec` bypasses `IsValidNodeType`/`IsValidProfile` while `CreateNodeSpec` still rejects them.
- [ ] `NewSubstrateSafetyCheckWorkflow` omits `CheckHostProfileStep`; the hardware step attaches resolution hints.
- [ ] No non-test consumer of `config.Get().Profile` runs in the substrate-only cluster-install path.
- [ ] `--skip-hardware-checks` still bypasses the substrate floor at cluster install.

## Tests run

```bash
task lint                       # 0 issues
go test ./pkg/hardware/...      # 95 passed (incl. new substrate-bypass test)
GOOS=linux go build ./...       # changed pkgs compile
```

Full unit + integration suites run in the UTM VM (Linux-only `internal/mount`):

```bash
task vm:test:unit
task vm:test:integration TEST_NAME='^Test_ClusterSetup$'
```

## Manual UAT

Run inside the UTM VM (Ubuntu), as a freshly-provisioned host.

### UAT-1 — Cluster install needs no profile/node-type

```bash
sudo ./bin/solo-provisioner-linux-amd64 kube cluster install --non-interactive
```

Expected: install proceeds through **Preflight Checks** (privileges, weaver user, then the four per-resource hardware steps — "Validating OS/CPU/memory/storage requirements"), **System Setup**, then **Kubernetes Setup**, ending:

```
Successfully installed Kubernetes Cluster
```

```bash
sudo kubectl get nodes    # node reports Ready
```

### UAT-2 — Flags hidden from help

```bash
./bin/solo-provisioner-linux-amd64 kube cluster install --help
```

Expected: **no** `--profile` / `-p` and **no** `--node-type` / `-n` in the flag list; only the error-control flags (`--stop-on-error`, `--rollback-on-error`, `--continue-on-error`) and inherited persistent flags.

### UAT-2b — Deprecated flags still accepted but ignored (backward compatibility)

The flags are retired the same way this CLI retires others (hidden + still parseable), so existing scripts do not break. Passing them must not error, must not change behavior, and must print a notice.

```bash
sudo ./bin/solo-provisioner-linux-amd64 kube cluster install --profile testnet --node-type block --non-interactive
```

Expected:
- Does **not** error (no `unknown flag`).
- Prints a WARN notice for each provided flag, e.g. `--profile is deprecated and ignored for 'kube cluster install': it validates only the Kubernetes substrate floor. ...`
- Runs the **same** substrate-only preflight as UAT-1 (the `testnet` value does **not** raise the floor to the block-node testnet floor) and completes with `Successfully installed Kubernetes Cluster`.

```bash
# Inherited --profile is likewise accepted (and silently ignored, as before) on uninstall:
sudo ./bin/solo-provisioner-linux-amd64 kube cluster uninstall --profile testnet
```

Expected: no `unknown flag` error; teardown proceeds normally.

### UAT-3 — Substrate floor still enforced

On a host below the substrate floor, set in UTM VM settings — easiest is **1 vCPU** (< 2 cores) or **1 GB RAM** (< 2 GB). Note the disk check reads *total physical capacity*, not free space, so a 64 GiB dev VM always clears the 20 GB floor regardless of usage.

```bash
sudo ./bin/solo-provisioner-linux-amd64 kube cluster install --non-interactive
```

Expected: fails at the matching per-resource step (e.g. "Validating CPU requirements") with the shortfall, `Set by: Kubernetes control-plane minimum`, and a resolution hint naming the minimum (e.g. "at least 2 CPU cores") plus the `--skip-hardware-checks` escape hatch.

### UAT-3b — Substrate preflight is four independent per-resource steps

Verifies the review change: substrate preflight reuses the node preflight's four per-resource steps rather than one combined step, so each resource has its own TUI entry, failure report, and resolution hint.

1. In UTM, set the VM below a single floor — e.g. **1 vCPU** (keep RAM at 2 GB+ so only CPU is short):

```bash
sudo ./bin/solo-provisioner-linux-amd64 kube cluster install --non-interactive
```

Expected:
- Preflight runs as **distinct per-resource steps** in order — `Validating OS requirements` (success), then `Validating CPU requirements` (failed) — not a single opaque "substrate hardware" step.
- The CPU failure carries its own message (`CPU does not meet K8s-Substrate Node requirements (minimum 2 cores)`), `Set by: Kubernetes control-plane minimum`, and a CPU-specific resolution hint.

2. To confirm each resource is reported independently, make a *different* single resource short — restore **2 vCPU** and set **1 GB RAM** — and re-run. Now `Validating CPU requirements` passes and `Validating memory requirements` fails with its own message and hint.

**Note on multiple simultaneous shortfalls:** the preflight stops at the **first** failing hardware step (e.g. CPU), so a host short on both CPU and memory surfaces only the CPU failure — matching the existing node preflight. This holds even with `--continue-on-error`: that flag only applies at the outer workflow level (letting it proceed into System Setup after the preflight fails); the nested preflight itself runs stop-on-error, because automa propagates execution mode only one level deep. Reporting all hardware shortfalls in a single run would be a separate, framework-level enhancement affecting every preflight.

### UAT-4 — Responsibility boundary moved (the point of the change)

On a host that meets the substrate floor but **not** the block-node testnet floor (e.g. 4 CPU / 8 GB / 100 GB disk):

```bash
# 1. Cluster install now succeeds (previously rejected on storage):
sudo ./bin/solo-provisioner-linux-amd64 kube cluster install --non-interactive
# -> Successfully installed Kubernetes Cluster

# 2. Workload validation still catches the mismatch, attributed to the workload:
sudo ./bin/solo-provisioner-linux-amd64 block node check --profile testnet
# -> fails on CPU/memory/storage with the block-node testnet floor and a
#    'Set by: block node testnet ...' attribution
```

### UAT-5 — block node install unchanged

On a host that meets the block-node testnet floor, with no existing cluster:

```bash
sudo ./bin/solo-provisioner-linux-amd64 block node install --profile testnet --non-interactive
```

Expected: runs the block-node **Preflight Checks** (workload-sized) → **System Setup** → **Kubernetes Setup** → block-node deployment, in that order — identical to before this PR.

## Risks

- `config.SetProfile` is no longer called at cluster install. Re-grep confirms the only `config.Get().Profile` readers (block-node prompt, alloy steps) do not run in the substrate-only `KubernetesSetupWorkflow` path.
- `block node install` recomposition is the main behavioral risk surface; `NodeSetupWorkflow` is unchanged and only the Kubernetes step was relocated, keeping ordering identical. Covered by the `Test_ClusterSetup` integration test and UAT-5.
- No embedded-template or install-time-config changes; no upgrade/migration path is touched.

### DIFF
--- a/cmd/cli/commands/kube/cluster/install.go
+++ b/cmd/cli/commands/kube/cluster/install.go
@@ -7,9 +7,6 @@ import (
 	"github.com/hashgraph/solo-weaver/cmd/cli/commands/common"
 	"github.com/hashgraph/solo-weaver/internal/rsl"
 	"github.com/hashgraph/solo-weaver/internal/workflows"
-	"github.com/hashgraph/solo-weaver/pkg/config"
-	"github.com/hashgraph/solo-weaver/pkg/hardware"
-	"github.com/hashgraph/solo-weaver/pkg/models"
 	"github.com/joomcode/errorx"
 	"github.com/spf13/cobra"
 )
@@ -19,28 +16,18 @@ var installCmd = &cobra.Command{
 	Short: "Install a Kubernetes Cluster",
 	Long:  "Run safety checks, setup a K8s cluster",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		flagProfile, err := common.FlagProfile().Value(cmd, args)
-		if err != nil {
-			return errorx.IllegalArgument.Wrap(err, "failed to get profile flag")
+		// Cluster install is workload-agnostic: it validates only the Kubernetes
+		// substrate hardware floor, so no --profile or --node-type is required.
+		// Both flags are deprecated: still accepted (hidden) for backward compatibility
+		// but ignored. Tell the operator if either was explicitly provided.
+		if cmd.Flags().Changed(common.FlagProfile().Name) {
+			logx.As().Warn().Msgf("--%s is deprecated and ignored for 'kube cluster install': it validates only the Kubernetes substrate floor. Profile-based sizing applies to 'block node' commands.",
+				common.FlagProfile().Name)
 		}
-
-		if flagProfile == "" {
-			return errorx.IllegalArgument.New("profile flag is required")
+		if cmd.Flags().Changed(common.FlagNodeType().Name) {
+			logx.As().Warn().Msgf("--%s is deprecated and ignored for 'kube cluster install': it validates only the Kubernetes substrate floor.",
+				common.FlagNodeType().Name)
 		}
-
-		// Validate node type and profile early for better error messages
-		if !hardware.IsValidNodeType(flagNodeType) {
-			return errorx.IllegalArgument.New("unsupported node type: %q. Supported types: %v",
-				flagNodeType, hardware.SupportedNodeTypes())
-		}
-
-		if !hardware.IsValidProfile(flagProfile) {
-			return errorx.IllegalArgument.New("unsupported profile: %q. Supported profiles: %v",
-				flagProfile, models.SupportedProfiles())
-		}
-
-		// Set the profile in the global config so other components can access it
-		config.SetProfile(flagProfile)
 
 		execMode, err := common.GetExecutionMode(flagContinueOnError, flagStopOnError, flagRollbackOnError)
 		if err != nil {
@@ -70,7 +57,7 @@ var installCmd = &cobra.Command{
 			return errorx.IllegalArgument.New("expected MachineRuntime to be *rsl.MachineRuntimeResolver but got %T", sr.Runtime.MachineRuntime)
 		}
 
-		wb := workflows.WithWorkflowExecutionMode(workflows.InstallClusterWorkflow(flagNodeType, flagProfile, "", skipHardwareChecks, mr), opts)
+		wb := workflows.WithWorkflowExecutionMode(workflows.InstallClusterWorkflow(skipHardwareChecks, mr), opts)
 		if err := common.RunWorkflowBuilder(cmd.Context(), wb); err != nil {
 			return err
 		}
@@ -81,7 +68,9 @@ var installCmd = &cobra.Command{
 }
 
 func init() {
-	common.FlagNodeType().SetVarP(installCmd, &flagNodeType, true)
+	// Deprecated: --node-type no longer affects cluster install (substrate-only floor).
+	// Kept hidden for backward compatibility; the value is ignored (see RunE notice).
+	common.FlagNodeType().SetVarPHidden(installCmd, &flagNodeType, false)
 	common.FlagStopOnError().SetVarP(installCmd, &flagStopOnError, false)
 	common.FlagRollbackOnError().SetVarP(installCmd, &flagRollbackOnError, false)
 	common.FlagContinueOnError().SetVarP(installCmd, &flagContinueOnError, false)
--- a/internal/workflows/preflight.go
+++ b/internal/workflows/preflight.go
@@ -414,3 +414,104 @@ func NewNodeSafetyCheckWorkflow(spec hardware.DeploymentSpec, skipHardwareChecks
 			notify.As().PhaseCompletion(ctx, stp, rpt, "Preflight Checks")
 		})
 }
+
+// CheckSubstrateHardwareStep validates the Kubernetes substrate hardware floor
+// (OS, CPU, memory, storage) in a single step, independent of any workload node
+// type or profile. It resolves requirements from the "k8s-substrate" provider via
+// hardware.CreateSubstrateSpec, which bypasses the node-type/profile validation gates.
+func CheckSubstrateHardwareStep() automa.Builder {
+	return automa.NewStepBuilder().WithId("validate-substrate-hardware").
+		WithExecute(func(ctx context.Context, stp automa.Step) *automa.Report {
+			hostProfile := hardware.GetHostProfile()
+			substrateSpec, err := hardware.CreateSubstrateSpec(hostProfile)
+			if err != nil {
+				return automa.FailureReport(stp, automa.WithError(
+					errorx.Decorate(err, "failed to resolve Kubernetes substrate requirements").
+						WithProperty(doctor.ErrPropertyResolution, []string{
+							"This is an internal error: verify the \"k8s-substrate\" requirements provider is registered.",
+						})))
+			}
+
+			reqs := substrateSpec.GetBaselineRequirements()
+			logx.As().Info().Msgf("substrate floor — required: %d cores, %d GB RAM, %d GB disk, OS %v",
+				reqs.MinCpuCores, reqs.MinMemoryGB, reqs.MinStorageGB, reqs.MinSupportedOS)
+
+			// Resolve the why-map once for operator-facing attribution on failure.
+			whyMap := map[string]string{}
+			if p, ok := hardware.Providers()[hardware.NodeTypeSubstrate]; ok {
+				if _, w, e := p.ComputeWithWhy(hardware.DeploymentSpec{NodeType: hardware.NodeTypeSubstrate}); e == nil {
+					whyMap = w
+				}
+			}
+
+			resolution := []string{
+				fmt.Sprintf("Provision a host that meets the Kubernetes control-plane minimum: %d CPU cores, %d GB RAM, %d GB free disk.",
+					reqs.MinCpuCores, reqs.MinMemoryGB, reqs.MinStorageGB),
+				"Re-run with --skip-hardware-checks to bypass substrate validation (not recommended).",
+			}
+
+			checks := []struct {
+				validate func() error
+				why      string
+				what     string
+			}{
+				{substrateSpec.ValidateOS, "", "OS"},
+				{substrateSpec.ValidateCPU, whyMap["cpu"], "CPU"},
+				{substrateSpec.ValidateMemory, whyMap["memory"], "memory"},
+				{substrateSpec.ValidateStorage, whyMap["storage"], "storage"},
+			}
+			for _, c := range checks {
+				if err := c.validate(); err != nil {
+					baseErr := errorx.IllegalState.Wrap(err, "%s validation failed for Kubernetes substrate", c.what).
+						WithProperty(doctor.ErrPropertyResolution, resolution)
+					if c.why != "" {
+						baseErr = baseErr.WithProperty(models.ErrPropertyWhyFloor, c.why)
+					}
+					return automa.FailureReport(stp, automa.WithError(baseErr))
+				}
+			}
+			return automa.SuccessReport(stp)
+		}).
+		WithPrepare(func(ctx context.Context, stp automa.Step) (context.Context, error) {
+			notify.As().StepStart(ctx, stp, "Validating Kubernetes substrate hardware")
+			return ctx, nil
+		}).
+		WithOnFailure(func(ctx context.Context, stp automa.Step, rpt *automa.Report) {
+			notify.As().StepFailure(ctx, stp, rpt, "Validating Kubernetes substrate hardware")
+		}).
+		WithOnCompletion(func(ctx context.Context, stp automa.Step, rpt *automa.Report) {
+			notify.As().StepCompletion(ctx, stp, rpt, "Validating Kubernetes substrate hardware")
+		})
+}
+
+// NewSubstrateSafetyCheckWorkflow creates a safety check workflow for the Kubernetes
+// substrate — the hardware floor Kubernetes itself needs, independent of any workload
+// node type or profile. Unlike NewNodeSafetyCheckWorkflow it omits CheckHostProfileStep
+// (there is no node type / profile to validate) and validates hardware via the single
+// CheckSubstrateHardwareStep. If skipHardwareChecks is true, hardware validation is excluded.
+func NewSubstrateSafetyCheckWorkflow(skipHardwareChecks bool) *automa.WorkflowBuilder {
+	preflightSteps := []automa.Builder{
+		CheckPrivilegesStep(),
+		CheckWeaverUserStep(),
+	}
+
+	if skipHardwareChecks {
+		logx.As().Warn().Msg("Substrate hardware validation (OS, CPU, memory, storage) will be skipped due to --skip-hardware-checks flag")
+	} else {
+		preflightSteps = append(preflightSteps, CheckSubstrateHardwareStep())
+	}
+
+	return automa.NewWorkflowBuilder().
+		WithId("substrate-preflight").
+		Steps(preflightSteps...).
+		WithPrepare(func(ctx context.Context, stp automa.Step) (context.Context, error) {
+			notify.As().PhaseStart(ctx, stp, "Preflight Checks")
+			return ctx, nil
+		}).
+		WithOnFailure(func(ctx context.Context, stp automa.Step, rpt *automa.Report) {
+			notify.As().PhaseFailure(ctx, stp, rpt, "Preflight Checks")
+		}).
+		WithOnCompletion(func(ctx context.Context, stp automa.Step, rpt *automa.Report) {
+			notify.As().PhaseCompletion(ctx, stp, rpt, "Preflight Checks")
+		})
+}
--- a/internal/workflows/preflight.go
+++ b/internal/workflows/preflight.go
@@ -20,8 +20,14 @@ import (
 	"github.com/joomcode/errorx"
 )
 
-// createNodeSpec creates the appropriate node spec based on a DeploymentSpec and host profile
+// createNodeSpec creates the appropriate node spec based on a DeploymentSpec and host profile.
+// The Kubernetes substrate is resolved via CreateSubstrateSpec, which bypasses the
+// node-type/profile validation gates (the substrate is not a user-facing node type and
+// carries no profile); everything else goes through the standard CreateNodeSpec path.
 func createNodeSpec(spec hardware.DeploymentSpec, hostProfile hardware.HostProfile) (hardware.Spec, error) {
+	if spec.NodeType == hardware.NodeTypeSubstrate {
+		return hardware.CreateSubstrateSpec(hostProfile)
+	}
 	return hardware.CreateNodeSpec(spec, hostProfile)
 }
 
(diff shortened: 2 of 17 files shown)


internal/workflows/preflight.go
```diff
@@ -414,3 +414,104 @@ func NewNodeSafetyCheckWorkflow(spec hardware.DeploymentSpec, skipHardwareChecks
 			notify.As().PhaseCompletion(ctx, stp, rpt, "Preflight Checks")
 		})
 }
+
+// CheckSubstrateHardwareStep validates the Kubernetes substrate hardware floor
+// (OS, CPU, memory, storage) in a single step, independent of any workload node
+// type or profile. It resolves requirements from the "k8s-substrate" provider via
+// hardware.CreateSubstrateSpec, which bypasses the node-type/profile validation gates.
+func CheckSubstrateHardwareStep() automa.Builder {
```
leninmehedy: CheckSubstrateHardwareStep — all four hardware checks collapsed into one step, only the first failure surfaces

The established pattern for hardware validation is four independent steps — CheckOSStep, CheckCPUStep, CheckMemoryStep, CheckStorageStep — each with its own TUI entry and failure report. 

Can you check if a host is short on both memory and storage, whether both errors surfaces or not

internal/workflows/preflight.go
```diff
@@ -414,3 +414,104 @@ func NewNodeSafetyCheckWorkflow(spec hardware.DeploymentSpec, skipHardwareChecks
...
+				return automa.FailureReport(stp, automa.WithError(
+					errorx.Decorate(err, "failed to resolve Kubernetes substrate requirements").
+						WithProperty(doctor.ErrPropertyResolution, []string{
+							"This is an internal error: verify the \"k8s-substrate\" requirements provider is registered.",
+						})))
+			}
+
+			reqs := substrateSpec.GetBaselineRequirements()
+			logx.As().Info().Msgf("substrate floor — required: %d cores, %d GB RAM, %d GB disk, OS %v",
+				reqs.MinCpuCores, reqs.MinMemoryGB, reqs.MinStorageGB, reqs.MinSupportedOS)
```
leninmehedy: During the `whyMap` creation,`hardware.Providers()[hardware.NodeTypeSubstrate]` already returns `req` which is ignored. So we can just use that instead of creating a custom `CreateSubstrateSpec` which seems to be redundant.

I am ok with how it is, but just added the comment to validate.

cmd/cli/commands/kube/cluster/install.go
```diff
@@ -81,7 +68,9 @@ var installCmd = &cobra.Command{
 }
 
 func init() {
-	common.FlagNodeType().SetVarP(installCmd, &flagNodeType, true)
+	// Deprecated: --node-type no longer affects cluster install (substrate-only floor).
+	// Kept hidden for backward compatibility; the value is ignored (see RunE notice).
+	common.FlagNodeType().SetVarPHidden(installCmd, &flagNodeType, false)
```
leninmehedy, line 73: Claude gave a review as below, please verify before implementing;

----
cmd/cli/commands/kube/cluster/install.go — cmd.Flags().Changed("profile") will silently miss the deprecation warning for --profile

--profile is registered as a PersistentFlag on the parent kubeCmd (not on installCmd itself). The established project convention for detecting whether an inherited persistent flag was explicitly set is cmd.InheritedFlags().Lookup("profile").Changed, as used at cmd/cli/commands/block/node/init.go. Using cmd.Flags().Changed("profile") in a child command for a parent-owned persistent flag may return false even when the user explicitly passed --profile, silently suppressing the deprecation notice. The --node-type check on the next line (cmd.Flags().Changed(common.FlagNodeType().Name)) is fine because that flag is registered on installCmd itself via SetVarPHidden.

Fix: Replace with cmd.InheritedFlags().Lookup(common.FlagProfile().Name) and check .Changed on the returned *pflag.Flag, matching the existing project pattern.

---

leninmehedy (changes requested):

cmd/cli/commands/kube/cluster/install.go
alex-au: Interesting, the `--profile` warning did fire in my UAT test case 2b, both warnings appeared:
```
WRN --profile is deprecated and ignored for 'kube cluster install'...
WRN --node-type is deprecated and ignored for 'kube cluster install'...
```
cobra merges inherited persistent flags into `cmd.Flags()` before `RunE`, so `Changed` is tracked on the shared flag. `InheritedFlags().Lookup("profile")` is used to mutate `Changed` during config resolution, so `cmd.Flags().Changed` is not unreliable as Claude commented.

internal/workflows/preflight.go
alex-au: Great point, the 4-step change resolved this one as well now. The provider only returns the raw `BaselineRequirements` numbers. the preflight steps need a `Spec` with the `Validate*` methods to check them against the host. And `CreateSubstrateSpec` supplies that, so it's no longer a redundant helper now.

pkg/hardware/factory.go
```diff
@@ -67,3 +67,20 @@ func CreateNodeSpec(spec DeploymentSpec, hostProfile HostProfile) (Spec, error)
...
+
+// CreateSubstrateSpec creates a spec for the Kubernetes substrate floor.
+//
+// Unlike CreateNodeSpec, it deliberately bypasses the IsValidNodeType / IsValidProfile
+// gates: the substrate is not a user-facing node type and carries no profile. It resolves
+// requirements straight from the "k8s-substrate" provider via NewNodeSpec, which only
+// consults the provider registry. The substrate provider ignores profile and options.
+func CreateSubstrateSpec(hostProfile HostProfile) (Spec, error) {
+	spec := DeploymentSpec{NodeType: NodeTypeSubstrate}
+
```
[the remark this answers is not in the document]
alex-au: Good catch, fixed in [node_spec.go](https://github.com/hashgraph/solo-weaver/pull/807/changes/0d572b27ce8ac2ebf2fb2eb49ff414d5e952b8b8#diff-887264f947a9a0a66b8e579039401f7b7139b1108b2bf2e0229b45955d3154e5)

internal/workflows/preflight.go
alex-au: Good catch, I switched it to the 4 step approach. `NewSubstrateSafetyCheckWorkflow` now reuses the existing `CheckOSStep / CheckCPUStep / CheckMemoryStep / CheckStorageStep` with a substrate spec, routed through `CreateSubstrateSpec`. So each resource gets its own TUI entry, failure report, and resolution hint to match the node preflight structure. `CheckHostProfileStep` is omitted since there's no node type/profile to validate.

To the question about "do both surface"? I dug into this and the answer is no, it turns out the same is true of the existing node preflight. The preflight is a nested sub-workflow, and automa only propagates execution mode one level deep ([workflow_builder.go](https://github.com/automa-saga/automa/blob/main/workflow_builder.go#L93)). So the preflight runs stop-on-error regardless and halts at the first failing hardware step. Therefore, a host short on both CPU and memory reports only CPU.

So the 4-step split delivers per-resource granularity (distinct entries + per-resource hints), consistent with the node preflight, but not simultaneous multi-failure reporting. Surfacing all hardware shortfalls in one run would be a separate, framework level change affecting every preflight. Happy to file a follow up issue if that's worth doing.

leninmehedy (approved):
Awesome work! Thanks

internal/workflows/preflight.go
leninmehedy: Let's keep as it is with "stop-on-error" behaviour.

PR #809, opened 2026-07-07
feat(tui): navigable back/forward block node install wizard

### LINKED ISSUE #582: TUI Installation Wizard — Full Back/Forward Navigation
## Context

All block-node prompts are sequenced in `promptForMissingFlags` in `cmd/weaver/commands/block/node/init.go` as four sequential `Run*Prompts` calls. Each call creates and runs its own independent `huh.Form`, so once the user leaves a page they can never go back. The user wants all prompt pages navigable as a single installation wizard where Shift+Tab goes back and Enter/Tab goes forward.

`charmbracelet/huh` natively supports this when all groups are passed to a single `huh.NewForm(group1, group2, ...)` — each group is one wizard page and keyboard navigation between pages is built-in (shown via `WithShowHelp(true)`, which is already set).

---

## Wizard Page Structure

For a typical `block-node install` run, the wizard presents up to 6 pages:

| Page | Content | Condition |
|------|---------|-----------|
| 1 | Profile select | skipped if `--profile` set |
| 2 | Namespace / Release Name / Chart Version / Retention | skipped if all flags set |
| 3 | Storage Path Mode | skipped if any storage flag set |
| 4 | Path inputs (base path OR 5 individual paths) | dynamic based on page 3 |
| 5 | Plugin Preset | skipped if `--plugin-preset` or `--plugins` set |
| 6 | Custom Plugin Multi-select | conditional: only when Custom chosen on page 5 |

Pages 4 and 6 use `huh.Field.WithHideFunc` / `huh.Group.WithHideFunc` to dynamically show/hide based on the user's selection on the preceding page — including when the user navigates backward and changes their choice.

---

## Architecture: `Wizard` type

Introduce a `Wizard` struct in `prompt.go` that accumulates groups and post-run callbacks, then executes one `huh.NewForm()`:

```go
type Wizard struct {
    groups   []*huh.Group
    afterRun []func()   // collect cv entries after form.Run() returns
}

func NewWizard() *Wizard

// Run executes all accumulated groups in a single huh form, then fires afterRun callbacks.
func (w *Wizard) Run(cv *ChosenValues) error
```

`addGroups` (unexported) appends groups + callback to the wizard.

---

alex-au:
## Description

Makes the interactive `block node install` prompts a single navigable wizard: **Shift+Tab** goes back, **Tab/Enter** goes forward, across every page. Previously `promptForMissingFlags` ran up to six independent `huh.NewForm(...).Run()` calls in sequence, each committing before the next began — so once you left a page you could never return, and correcting any earlier answer meant Ctrl+C and re-entering everything from scratch.

The fix introduces a `Wizard` type that accumulates `huh` groups (one per page) plus post-run callbacks and runs them in one `huh.NewForm(...)`, where huh's built-in back/forward navigation works natively. The two conditional pages — storage **single-base-path vs individual-paths**, and the **custom plugin multi-select** — use `huh.Group.WithHideFunc` so they show/hide dynamically as the controlling selection changes, including when navigating backward and changing it. Post-processing (recording chosen values, joining the custom plugin list, clearing the unused storage-mode's targets) moves into `afterRun` callbacks because a field's value is only final once the shared form completes.

The old `RunSelectPrompts` / `RunInputPrompts` / `RunStoragePathPrompts` / `RunPluginPresetPrompts` functions are retained as thin single-page wrappers, so the daemon prompt flow and existing unit tests are unchanged.

One design constraint: `huh` supports hiding only at the **group** level, not per field. So the individual-path optional-storage set and the custom plugin option list are fixed at construction from the effective chart version — editing chart-version later in the same wizard does not re-shape those fields. This is a rare-edge behavior change (change chart-version *and* use individual paths/custom plugins *and* cross an optional-storage boundary in one interactive run); the core requirement (storage/plugin pages reacting to their controlling page on back-nav) is fully met.

### Files changed

| File | Change |
|---|---|
| `internal/ui/prompt/prompt.go` | Add `Wizard` type (`NewWizard`, `addGroups`, `Run`) and `AddSelectPrompts`/`AddInputPrompts`; convert `RunSelectPrompts`/`RunInputPrompts` to one-page wrappers |
| `internal/ui/prompt/blocknode.go` | Add `AddStoragePathPrompts` (mode select + mode-gated base/individual groups, validator gating, target normalisation) and `AddPluginPresetPrompts` (preset + conditional custom group); convert the two `Run*` functions to wrappers |
| `cmd/cli/commands/block/node/init.go` | Rewrite `promptForMissingFlags` to build+run one `prompt.Wizard` |
| `internal/ui/prompt/wizard_test.go` | New white-box tests for group accumulation, skip-gating, and `afterRun` target normalisation |

### Review guide

**Checklist**
- `prompt.go` — `Wizard.Run` is a no-op with zero groups (`prompt.go`, `Run`), preserving the "all flags provided / non-interactive" bypass. `Add*` still honor the skip rules (`flagWasSet`, unregistered-flag skip) and pre-fill `EffectiveValue`.
- `blocknode.go` `AddStoragePathPrompts` — both storage-mode variants are pre-built as separate groups gated by `WithHideFunc`; `storagePathField` suppresses a field's validator while its mode is inactive (so a hidden, pre-filled path can't block submit); `afterRun` clears the **unused** mode's targets so downstream storage-mode inference in `prepareBlocknodeInputs` stays unambiguous.
- `blocknode.go` `AddPluginPresetPrompts` — the custom multi-select group is hidden unless preset == Custom; its `Validate` short-circuits when not Custom; `afterRun` sets `--plugins` only for Custom and **clears** it otherwise (no contamination from a pre-selected-but-hidden list).
- `init.go` — pages added in order (inputs before storage/plugins so `flagChartVersion` is pre-filled first); profile propagation to the inherited persistent flag moved to after `w.Run()`.

**Tests**
```bash
go test -race -tags='!integration' ./internal/ui/prompt/...
GOOS=linux GOARCH=arm64 golangci-lint run ./internal/ui/prompt/... ./cmd/cli/commands/block/node/...
```

**Manual UAT** — `sudo solo-provisioner block node install` (no flags, on a TTY), then:
- Shift+Tab from any later page returns to the previous page; entered values are preserved.
- On the Storage Path Mode page, switch Single↔Individual and go forward — the path page re-renders to match, even after backward nav.
- Pick the Custom plugin preset → the multi-select page appears; go back and pick a canned preset → it disappears and `--plugins` ends empty.
- No correction requires Ctrl+C; the printed "Selected Inputs" summary matches the final choices.
- Regression: `--non-interactive` and fully-flagged runs bypass all prompts.

**Risks / rollback** — Pure prompt-layer change; no state/schema/workflow impact. Rollback = restore the four sequential `Run*` calls in `promptForMissingFlags`. Behavior note called out above: optional-storage/custom-plugin field sets are pinned to the effective chart version at wizard construction.

### Related Issues

* Closes #582


leninmehedy (approved):
I haven't tested, but LGTM. Thanks for improving UX of the tool.

alex-au:
> I haven't tested, but LGTM. Thanks for improving UX of the tool.

Thanks Lenin, I UAT tested and it works for me, so I'll merge it 😀

PR #813, opened 2026-07-08
fix(daemon): harden daemon concurrency

### LINKED ISSUE #697: fix(daemon): harden two concurrency edges (untracked probe goroutine + latent soak WaitGroup race)
## Summary

Two low-severity concurrency hardening items in the daemon, surfaced during review of #688. Neither is a correctness bug in practice (no data race, no panic under normal operator cadence), so they were intentionally **not** blocked on for #688. Filing as a single follow-up so they aren't lost.

## 1. `runComponentProbes` runs outside the errgroup → non-deterministic shutdown

`internal/daemon/daemon.go:350`

```go
eg, ctx := errgroup.WithContext(ctx)
eg.Go(func() error { return d.server.Start(ctx) })
eg.Go(func() error { return d.componentSupervisor(ctx) })

go d.runComponentProbes(ctx)   // <-- not tracked by the errgroup

return eg.Wait()
```

`Run` returns when `eg.Wait()` completes (server + supervisor only). The probe goroutine can still be mid-cycle, and its post-`Run` write to `d.probeErrors` (`daemon.go:237,247`) is therefore unsynchronized w.r.t. `Run` returning.

**Why it's not urgent:** the write target is an `atomic.Pointer`, so there is **no data race**; the inner per-probe goroutines (`daemon.go:196`) are tracked by a local `wg.Wait()` so they don't leak past one cycle; and the loop is one-shot — once all probes pass it returns (`daemon.go:245`), so in the healthy steady state the goroutine is already gone. The straggler only exists while prerequisites are still failing, and it exits on `ctx.Done()` between cycles (`daemon.go:250`). Impact is purely non-deterministic shutdown ordering.

**Fix:** track it in the errgroup (or a dedicated `WaitGroup`) so `Run` returns only after the probe loop has stopped.

## 2. Latent `WaitGroup` Add-vs-Wait window between `TryStop` and the dispatch loop

`internal/daemon/consensus/migration_monitor.go`

`TryStop` calls `mm.soakWg.Wait()` (`:270`) after loading a non-nil `soakCancel` (`:265`). The dispatch loop calls `mm.soakWg.Add(1)` (`:306`) on each new `start`. There is a narrow interleaving where `Add` brings the counter `0→1` concurrently with `TryStop`'s `Wait`:

1. Soak running, counter=1, `soakCancel=&c`.
2. `TryStop` loads the non-nil `cancelPtr`.
3. The watcher fully exits: `soakCancel→nil` (`:330`), `soakActive→false` (`:346`), `Done()` → counter **0** (`:348`).
4. A fresh `start` is accepted (`soakActive` is now false) and dispatched: `Add(1)` **0→1** (`:306`) — concurrent with `TryStop` finally reaching `soakWg.Wait()` (`:270`).

This is the `sync.WaitGroup`-documented forbidden pattern (positive-delta-from-zero `Add` concurrent with `Wait`), which can panic *"WaitGroup is reused before previous Wait has returned."*

> Note: the path originally flagged during review — `TryStop` vs `resumeIfNeeded` — is **not** reachable. `resumeIfNeeded` runs once before the dispatch loop, and its `Add(1)` (`:591`) completes before the spawned `run()` sets `soakCancel`, so `TryStop` always observes counter ≥1 there.

**Why it's not urgent:** soak `start`/`stop` are human-driven CLI commands, seconds apart, with a single-soak-in-flight invariant. The window is a few microseconds and requires `stop`, the soak ending, and a fresh `start` to interleave exactly. Effectively unreachable in production; a tight stop/start stress loop under `-race` could in theory trip it.

**Fix:** serialize soak `start`/`stop` under a mutex, or restructure so `soakWg.Wait()` is not called outside the dispatch goroutine (e.g. signal stop and let `Run`'s deferred `Wait` drain it).

## Acceptance criteria

- [ ] `runComponentProbes` is awaited before `Daemon.Run` returns.
- [ ] No `WaitGroup` `Add`-from-zero can race `Wait`; ideally demonstrated with a tight stop/start test run under `-race`.
- [ ] No behavior change to the happy path; existing daemon tests still pass.

_Surfaced during review of #688 (daemon core, closes #499)._

radkomih:
## Description

- Adds a soakLifecycleMu mutex to MigrationMonitor that serializes soakWg.Add (in Run/resumeIfNeeded) against soakWg.Wait (in TryStop), fixing a WaitGroup data race where a 0→1 Add could overlap a concurrent Wait.
- Moves runComponentProbes from a detached go statement into the Daemon.Run errgroup (returning nil so it never cancels the group), so Run no longer returns while the component-probe loop is still in-flight.
- Adds Test_MigrationMonitor_ConcurrentStartStop, which hammers TryEnqueue/TryStop from two goroutines to exercise the Add/Wait race under -race.
- Adds TestRun_AwaitsProbeGoroutine, which uses a gate-blocked probe to prove Run waits for the probe loop to finish before returning.

### Related Issues

- Closes #697

### DIFF
--- a/internal/daemon/consensus/migration_monitor.go
+++ b/internal/daemon/consensus/migration_monitor.go
@@ -81,6 +81,9 @@ type MigrationMonitor struct {
 	// quiescence before returning.
 	soakWg sync.WaitGroup
 
+	// soakLifecycleMu serializes soakWg.Add (Run/resumeIfNeeded) against soakWg.Wait (TryStop)
+	soakLifecycleMu sync.Mutex
+
 	// soakCancel cancels the per-watcher context. Set atomically when a watcher
 	// goroutine starts; cleared when it exits. Used by TryStop.
 	soakCancel atomic.Pointer[context.CancelFunc]
@@ -281,7 +284,10 @@ func (mm *MigrationMonitor) TryStop(deleteState bool) bool {
 		return false
 	}
 	(*cancelPtr)()
+	// Serialize against the Add sites so Wait never overlaps a 0->1 Add
+	mm.soakLifecycleMu.Lock()
 	mm.soakWg.Wait()
+	mm.soakLifecycleMu.Unlock()
 	if deleteState && mm.stateFilePath != "" {
 		if err := os.Remove(mm.stateFilePath); err != nil && !os.IsNotExist(err) {
 			logx.As().Warn().Err(err).
@@ -315,9 +321,10 @@ func (mm *MigrationMonitor) Run(ctx context.Context) error {
 	for {
 		select {
 		case req := <-mm.soakStartCh:
-			// soakWg.Add before goroutine start so the deferred soakWg.Wait()
-			// above accounts for it.
+			// Add before spawn (accounted by the deferred Wait), locked so it can't race TryStop's Wait
+			mm.soakLifecycleMu.Lock()
 			mm.soakWg.Add(1)
+			mm.soakLifecycleMu.Unlock()
 			go mm.run(ctx, req, false)
 		case <-ctx.Done():
 			// Intentional: if soakStartCh has a buffered item at the same time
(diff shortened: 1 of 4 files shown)


alex-au (approved):
LGTM!

PR #820, opened 2026-07-09
fix(daemon): join monitor goroutine before TempDir cleanup

### LINKED ISSUE #819: Flaky `Test_MigrationMonitor_EmitsSoakStarted`: `Run` goroutine outlives `t.TempDir` cleanup
## Problem

`Test_MigrationMonitor_EmitsSoakStarted` (`internal/daemon/consensus/migration_monitor_test.go`) is flaky. It fails intermittently during temp-dir teardown, not on an assertion:

```
--- FAIL: Test_MigrationMonitor_EmitsSoakStarted (0.00s)
    testing.go:1464: TempDir RemoveAll cleanup: unlinkat /tmp/Test_MigrationMonitor_EmitsSoakStarted3981412625/002: directory not empty
FAIL	github.com/hashgraph/solo-weaver/internal/daemon/consensus
```

This surfaced as the sole red check on the unrelated dependabot PR #816 (`golang.org/x/text` bump), but it is **not** caused by that PR. Running the test on `main` 200× with `-race` reproduces it ~6% of the time (12/200 failures) with the identical message, so it flakes every PR that happens to lose the race.

Root cause: the test launches `mm.Run(ctx)` in a goroutine and returns as soon as `require.Eventually` observes the `SoakStarted` event. Teardown is only `defer cancel()`, which is asynchronous and never joins the goroutine. `Run` also writes `cutover-state.jsonl` into `stateDir := t.TempDir()` via `writeSoakState` (create `.tmp` → fsync → rename → dir fsync). When the test returns, Go's `t.TempDir` cleanup calls `RemoveAll` on `stateDir` while the `Run` goroutine is still mid-write creating the `.tmp` file, leaving the directory non-empty. `stateDir` is the second temp dir created, matching the `…/002` in the error.

## Proposed fix

Join the `Run` goroutine before temp-dir cleanup so it cannot outlive `RemoveAll`. Register the join via `t.Cleanup` *after* the `t.TempDir()` call so it runs before `TempDir`'s own removal:

```go
done := make(chan struct{})
go func() { defer close(done); _ = mm.Run(ctx) }()
t.Cleanup(func() { cancel(); <-done })
```

Apply to every test in the file that starts `mm.Run` in a goroutine against a `t.TempDir()`-backed `stateDir`.

## Acceptance

- [ ] `go test -count=200 -race -run Test_MigrationMonitor ./internal/daemon/consensus/` passes with zero failures
- [ ] No `Run` goroutine writes into a `stateDir` after its owning test has returned

alex-au:
## Description

`Test_MigrationMonitor_*` in `internal/daemon/consensus/migration_monitor_test.go` are flaky. Each launches `mm.Run(ctx)` in a background goroutine and returns the instant `require.Eventually` observes the expected event, with teardown being only `defer cancel()`. Because `cancel()` is asynchronous and never joins the goroutine, `Run` can still be writing `cutover-state.jsonl` into the `t.TempDir()` state dir when Go's `TempDir` cleanup calls `RemoveAll`, failing the test during teardown:

```
--- FAIL: Test_MigrationMonitor_EmitsSoakStarted (0.00s)
    testing.go:1464: TempDir RemoveAll cleanup: unlinkat /tmp/Test_MigrationMonitor_EmitsSoakStarted.../002: directory not empty
FAIL	github.com/hashgraph/solo-weaver/internal/daemon/consensus
```

This is a pre-existing flake on `main`, not caused by any dependency change. It surfaced as the sole red check on the unrelated dependabot PR #816 (`golang.org/x/text` bump), [Link](https://github.com/hashgraph/solo-weaver/actions/runs/28989612464/job/86026930666?pr=816) to more details of the unit test; re-running that job cleared it. Reproduced on `main` at ~6% (12/200 runs with `-race`).

The fix adds a `startMonitor(t, mm)` helper that registers a `t.Cleanup` which cancels the context and **waits for `Run` to return** (via a `done` channel) before the `t.TempDir` directories are removed. Since `t.Cleanup` functions run in reverse registration order, registering the join after the `t.TempDir()` calls guarantees the goroutine has stopped before `RemoveAll`. `newTestLogger` now also closes its logger via `t.Cleanup` (registered before the join) so the goroutine can't write to a closed logger. This mirrors the already-correct explicit join in `Test_MigrationMonitor_ConcurrentStartStop`. Test-only change; no production code touched.

### Verification

```
# Was ~6% fail on main; now 0 across 1800 runs
go test -count=200 -race -run Test_MigrationMonitor -tags='!integration' ./internal/daemon/consensus/
# Full package
go test -count=1 -race -tags='!integration' ./internal/daemon/consensus/   # 31 passed
```

### Related Issues

* Closes #819

### DIFF
--- a/internal/daemon/consensus/migration_monitor_test.go
+++ b/internal/daemon/consensus/migration_monitor_test.go
@@ -23,13 +23,32 @@ import (
 // --- Test helpers ---
 
 // newTestLogger creates a real EventLogger writing to a temp file and returns
-// the logger and the file path. The caller must close the logger when done.
+// the logger and the file path. The logger is closed via t.Cleanup, which runs
+// after any cleanup registered later (e.g. startMonitor's goroutine join), so
+// the monitor goroutine is guaranteed to have stopped writing before Close.
 func newTestLogger(t *testing.T) (*eventlog.EventLogger, string) {
 	t.Helper()
 	dir := t.TempDir()
 	logger, err := eventlog.NewAppend(dir, "test-events.jsonl")
 	require.NoError(t, err)
+	t.Cleanup(func() { _ = logger.Close() })
 	return logger, logger.Path()
+}
+
+// startMonitor runs mm.Run in a background goroutine and registers a cleanup
+// that cancels the context and waits for Run to return before the test's
+// t.TempDir directories are removed. The join is required: Run writes
+// cutover-state.jsonl into the state dir, and without waiting for it to exit the
+// goroutine can still be creating that file when TempDir's RemoveAll fires,
+// failing the test with "TempDir RemoveAll cleanup: ... directory not empty".
+// Registering the join with t.Cleanup after the t.TempDir() calls guarantees it
+// runs before their removal (cleanups run in reverse registration order).
+func startMonitor(t *testing.T, mm *consensus.MigrationMonitor) {
+	t.Helper()
+	ctx, cancel := context.WithCancel(context.Background())
+	done := make(chan struct{})
+	go func() { defer close(done); _ = mm.Run(ctx) }()
+	t.Cleanup(func() { cancel(); <-done })
 }
 
 // readEvents reads all JSON lines from the given JSONL file and returns them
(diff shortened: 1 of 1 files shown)


brunodam (approved):
LGTM

PR #827, opened 2026-07-10
fix(tui): react to edited chart version in block node install wizard

### LINKED ISSUE #826: Interactive `block node install` wizard ignores edited Chart Version when shaping storage-path and plugin pages
## Problem

In the interactive `block node install` wizard, editing the **Chart Version** input does not re-shape the pages that depend on it. The storage-path and plugin-selection pages are built from the *default* chart version captured at wizard-construction time, not the value the operator actually types.

Concretely: the storage default is `0.35.1` (`deps.BLOCK_NODE_VERSION`). If you change **Chart Version** to `0.37.1` in the wizard, the next page still shows the `0.35.1` optional-storage set:

```
Individual paths  (archive, live, log, verification, plugins must all be provided)
```

instead of the correct `0.37.x` set:

```
Individual paths  (archive, live, log, plugins, application-state must all be provided)
```

`verification` applies only to chart versions `< 0.37.0`; `application-state` applies only to `>= 0.37.0` (boundary `BlockNodeApplicationStateRequiredVersion = "0.37.0-0"`). Picking individual paths after editing the version therefore prompts for the wrong directories, and the resulting install/upgrade requests the wrong optional storages.

### Root cause

`promptForMissingFlags` builds one `huh` wizard up-front and passes `flagChartVersion` **by value** to the page builders:

- `cmd/cli/commands/block/node/init.go:152` → `AddStoragePathPrompts(w, cmd, defaults, flagChartVersion, ...)`
- `cmd/cli/commands/block/node/init.go:163` → `AddPluginPresetPrompts(w, cmd, defaults, ..., flagChartVersion, ...)`

`AddInputPrompts` pre-fills `flagChartVersion` to the *effective default* at build time (`internal/ui/prompt/prompt.go:302-305`), so the builders snapshot `0.35.1`. `AddStoragePathPrompts` then calls `blocknode.GetApplicableOptionalStorages(chartVersion)` once (`internal/ui/prompt/blocknode.go:345`) and bakes the field set + "…must all be provided" label into a static `huh` group. Edits made during `w.Run()` update the variable but cannot reshape the already-constructed group.

The version-gating logic itself (`GetApplicableOptionalStorages`, `PluginsForVersion`) is correct — only the snapshot is stale.

### Scope

- **Storage-paths page** — actively broken across the `0.37.0` optional-storage boundary (the reported case).
- **Plugin-selection page** — same defect, latent: `AddPluginPresetPrompts` → `PluginsForVersion(chartVersion)` builds the Custom multi-select options from the snapshot (`internal/ui/prompt/blocknode.go:627`). Not observable across `0.35.1 ↔ 0.37.1` (both `>= 0.35.0`, the plugin-history boundary), but breaks crossing `0.35.0`.
- **Not affected:** profile, namespace, release-name, retention (nothing downstream reads them) and `--values` (CLI-only, never edited in the wizard).

### Regression

Introduced by #809 (merge `8d3d95a`), which consolidated the previously sequential per-stage `huh.NewForm().Run()` calls into a single up-front wizard. Before #809, `RunInputPrompts` committed the chart-version edit *before* `RunStoragePathPrompts` was constructed, so the storage page read the final value. #809 documented this as a known "rare-edge behavior change"; in practice it is the exact flow operators hit when installing a `0.37.x` chart from a `0.35.1` default.

## Proposed fix

Make the storage-path and plugin pages evaluate applicability against the **live** chart-version value instead of a build-time snapshot. Pass a `*string` (or `func() string`) into `AddStoragePathPrompts` / `AddPluginPresetPrompts` and drive the dynamic content through `huh`'s per-render hooks:

- Gate each optional path (`verification`, `plugins`, `application-state`) in its own `WithHideFunc` group keyed to `GetApplicableOptionalStorages(*chartVersion)`.
- Recompute the "Individual paths (…)" label and the Custom-plugin options via `TitleFunc`/`OptionsFunc` bound to the live version.

## Acceptance

- [ ] With the default chart version `0.35.1`, changing **Chart Version** to a `0.37.x` value in the wizard makes the individual-paths page show `archive, live, log, plugins, application-state` (and hide `verification`).
- [ ] Changing back to a `< 0.37.0` version restores `verification` and hides `application-state`, including after backward (`Shift+Tab`) navigation.
- [ ] The Custom plugin multi-select reflects `PluginsForVersion` for the live chart version.
- [ ] Fully-flagged / `--non-interactive` runs still bypass all prompts (no regression to the #809 wizard).

alex-au:
## Description

The interactive `block node install` wizard did not react to a chart-version edit. Its storage-path and plugin pages were built from the *default* chart version (`0.35.1`, `deps.BLOCK_NODE_VERSION`) snapshotted at wizard-construction time. Changing **Chart Version** to a `0.37.x` value in the TUI still prompted for the pre-`0.37.0` optional-storage set — `archive, live, log, verification, plugins` — instead of the correct `0.37.x` set `archive, live, log, plugins, application-state`. Picking individual paths after the edit then prompted for the wrong directories and requested the wrong optional storages.

Root cause: `promptForMissingFlags` passed `flagChartVersion` **by value** to `AddStoragePathPrompts` / `AddPluginPresetPrompts`, which computed their field set and labels once (`GetApplicableOptionalStorages` / `PluginsForVersion`) and baked them into static `huh` groups. This was a regression from #809, which consolidated the previously sequential per-stage `huh` forms into a single up-front wizard — before #809 the chart-version form committed before the storage form was constructed, so the storage page read the final value.

The fix passes `&flagChartVersion` (by pointer) to both builders and drives their version-dependent content from the **live** value, so edits made on the earlier chart-version input page take effect as the operator navigates forward (full Shift+Tab / Tab navigation preserved). `huh` hides only whole groups (not individual fields), so each optional storage path now lives in its own hide-gated group.

This PR also bumps the default block-node chart version `deps.BLOCK_NODE_VERSION` from `0.35.1` to `0.37.1`. As a result, the wizard's default optional-storage set is now the `0.37.x` set (`application-state`, no `verification`); the reactivity fix above is what makes editing the version away from this new default behave correctly.

### Files changed

| File | Change |
|---|---|
| `cmd/cli/commands/block/node/init.go` | Pass `&flagChartVersion` to the storage and plugin builders; update the stale "pre-fill" comment. |
| `internal/ui/prompt/blocknode.go` | Storage page: one hide-gated group per optional storage whose `WithHideFunc` re-reads `*chartVersion`, dynamic mode-select label via `OptionsFunc`, and a version-aware `afterRun` that clears non-applicable optional paths. Plugin page: Custom multi-select options/warning via `OptionsFunc` / `DescriptionFunc` bound to `*chartVersion`, plus an `afterRun` selection filter. Wrappers `RunStoragePathPrompts` / `RunPluginPresetPrompts` adapted. |
| `internal/ui/prompt/wizard_test.go` | Updated call sites (pointer arg) and group count (now 6: mode + base + core + 3 optional groups); added `TestAddStoragePathPrompts_IndividualModeTracksChartVersionEdit` as the regression guard. |
| `pkg/deps/deps.go` | Bump default `BLOCK_NODE_VERSION` `0.35.1` → `0.37.1`. |

### Behavior note

Individual-paths mode now spans multiple pages (core `archive/live/log`, then one page each for the optional storages that apply to the live version) rather than a single page, because `huh` cannot hide individual fields within a group. Full back/forward navigation is retained. Optional-storage applicability is unchanged (`verification` `[0.26.2, 0.37.0)`, `plugins` `[0.28.1, ∞)`, `application-state` `[0.37.0, ∞)`).

### Review guide

**Checklist**
- `blocknode.go` `AddStoragePathPrompts` — `chartVersion *string`; `applicableOptional` re-reads `*chartVersion` per call; each optional path is its own `WithHideFunc` group gated on `!inIndividualMode() || !applicableOptional(name)`; `storagePathField`'s validator is suppressed while the field is not visible; mode-select label rebuilt via `OptionsFunc(modeOptions, chartVersion)` (option *values* stable, so selection is preserved across rebuilds); `afterRun` clears optional targets that do not apply to the final version.
- `blocknode.go` `AddPluginPresetPrompts` — `chartVersion *string`; Custom multi-select uses `OptionsFunc` / `DescriptionFunc` bound to `chartVersion`; `huh` reconciles the bound selection against rebuilt options (drops removed plugins); `afterRun` re-filters the selection against the final version and clears `--plugins` for non-custom presets.
- `init.go` — pointer passed to both builders; profile propagation after `w.Run()` unchanged.
- Pointer binding is required: `huh`'s `Eval` hashes the `bindings` argument, so a pointer is what lets the dynamic funcs observe later edits.

**Tests**
```bash
go test -race -tags='!integration' ./internal/ui/prompt/... ./internal/blocknode/...
task lint
```

**Manual UAT** — `sudo solo-provisioner block node install` on a TTY (no flags):
- On the chart-version page keep the new default (`0.37.1`) → forward → the storage mode page lists `Individual paths (archive, live, log, plugins, application-state must all be provided)`; the individual pages prompt `application-state` and not `verification`.
- Shift+Tab back to chart-version, change it to a `<0.37.0` value (e.g. `0.35.1`), forward again → the label now reads `… archive, live, log, verification, plugins …`; the individual pages prompt `verification` and not `application-state`.
- Change it back to a `0.37.x` value → `application-state` returns and `verification` disappears.
- Regression: `--non-interactive` and fully-flagged runs bypass all prompts; the printed "Selected Inputs" summary matches the final choices.

**Risks / rollback** — The prompt-layer change is pure (no state/schema/workflow/template impact). The `deps.BLOCK_NODE_VERSION` bump changes the chart version used by fresh `block node install` / `upgrade` runs that do not pass `--chart-version` (or set `blocknode.version` in config) from `0.35.1` to `0.37.1`; already-installed clusters are unaffected until an explicit upgrade. The compile-time default is not documented in `docs/quickstart.md`, so no doc change is required. Rollback = revert to the by-value snapshot and the single individual-paths group, and restore `BLOCK_NODE_VERSION`.

### Related Issues

* Closes #826

### DIFF
--- a/cmd/cli/commands/block/node/init.go
+++ b/cmd/cli/commands/block/node/init.go
@@ -122,9 +122,9 @@ func promptForMissingFlags(cmd *cobra.Command, args []string) error {
 	// Build a single wizard so every prompt page is navigable back and forward
 	// (Shift+Tab / Tab) instead of committing each stage in its own form. Pages are
 	// added in order; each Add* is a no-op when its flag(s) were already supplied on
-	// the command line. Input prompts must be added before the storage and plugin
-	// prompts because they pre-fill flagChartVersion, which the storage and plugin
-	// builders read to filter their options.
+	// the command line. The storage and plugin builders take &flagChartVersion (by
+	// pointer), so their version-dependent pages react to edits made on the earlier
+	// chart-version input page as the operator navigates forward.
 	w := prompt.NewWizard()
 
 	// profileTarget is a local copy the profile prompt writes to; after the wizard
@@ -149,7 +149,7 @@ func promptForMissingFlags(cmd *cobra.Command, args []string) error {
 	// Storage path prompts: mode select → conditional path inputs (single base
 	// path vs individual paths). Applied to all block node commands that configure
 	// storage (install, upgrade, reconfigure).
-	prompt.AddStoragePathPrompts(w, cmd, defaults, flagChartVersion, prompt.StoragePathTargets{
+	prompt.AddStoragePathPrompts(w, cmd, defaults, &flagChartVersion, prompt.StoragePathTargets{
 		BasePath:             &flagBasePath,
 		ArchivePath:          &flagArchivePath,
 		LivePath:             &flagLivePath,
@@ -160,7 +160,7 @@ func promptForMissingFlags(cmd *cobra.Command, args []string) error {
 	}, cv)
 
 	// Plugin preset prompt: preset select → conditional custom multi-select.
-	prompt.AddPluginPresetPrompts(w, cmd, defaults, &flagPluginPreset, &flagPlugins, flagChartVersion, flagValuesFile, cv)
+	prompt.AddPluginPresetPrompts(w, cmd, defaults, &flagPluginPreset, &flagPlugins, &flagChartVersion, flagValuesFile, cv)
 
 	// Run all accumulated pages as a single navigable form.
 	if err := w.Run(); err != nil {
--- a/internal/ui/prompt/blocknode.go
+++ b/internal/ui/prompt/blocknode.go
@@ -317,16 +317,18 @@ func storagePathField(p InputPrompt, active func() bool) huh.Field {
 //   - cv:           chosen-values collector for summary printing
 //
 // The optional-storage prompts (verification, plugins, application-state) are
-// gated on the registry's GetApplicableOptionalStorages(chartVersion) so the
-// operator only sees the storages the target chart actually needs. Because huh
-// cannot add or remove fields from a group mid-run, this applicable set is fixed
-// at construction from the effective chart version; editing the chart-version
-// input later in the same wizard does not re-shape the individual-path fields.
+// gated on the registry's RequiredByVersion for the *live* chart version. Each
+// optional path lives in its own huh group whose WithHideFunc re-reads
+// *chartVersion (huh can hide only whole groups, not individual fields), so
+// editing the chart-version input earlier in the same wizard re-shapes which
+// optional paths are shown — and the individual-paths label — when the operator
+// navigates forward. chartVersion is taken by pointer so huh's dynamic-func
+// bindings observe those later edits.
 func AddStoragePathPrompts(
 	w *Wizard,
 	cmd *cobra.Command,
 	defaults state.PromptDefaults,
-	chartVersion string,
+	chartVersion *string,
 	targets StoragePathTargets,
 	cv *ChosenValues,
 ) {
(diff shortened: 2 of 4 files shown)


Nana-EC (approved):
Conceptual approach seems right to me

brunodam (approved):
Great! Thank you!

PR #835, opened 2026-07-11
fix(tui): consolidate block node storage path prompts into one wizard page

### LINKED ISSUE #834: Consolidate individual storage path prompts into a single TUI page
## Problem

In the interactive `block node install` (and `upgrade`/`reconfigure`) wizard, the individual-paths storage configuration is split across **three separate TUI pages** instead of one:

1. Page 1 — `Archive Storage Path`, `Live Storage Path`, `Log Storage Path`
2. Page 2 — `Plugins Storage Path` only
3. Page 3 — `Application-State Storage Path` only

This is a side effect of how the wizard is built in `internal/ui/prompt/blocknode.go` (`AddStoragePathPrompts`): huh v1.0.0 can only hide whole `huh.Group`s, not individual fields, so each optional storage path (verification, plugins, application-state) was placed in its **own** group to let it appear/disappear reactively as the operator edits the chart version earlier in the wizard (PR #827 / issue #826). The core paths share one group, and each applicable optional path adds another — producing the multi-page layout.

Operators expect all the individual storage paths on a **single** page.

## Proposed fix

Consolidate the individual-paths inputs (archive, live, log + applicable optional paths) into a **single** `huh.Group`/page while preserving the per-field reactivity to chart-version edits. Because huh renders every field in a group and `Input.Skip()` is hardcoded `false`, introduce a small `huh.Field` wrapper that delegates to the embedded input but reports `Skip()==true` / `View()==""` when its hide-func says the optional path does not apply to the current chart version. huh re-evaluates field positions on every keypress, so dynamic `Skip()` keeps the reactive show/hide behavior intact.

## Acceptance

- [ ] The individual-paths mode renders archive, live, log, and every applicable optional storage path on a single wizard page.
- [ ] Editing the chart version earlier in the wizard still adds/removes optional paths on that single page (no regression of #826/#827).
- [ ] Optional paths that do not apply to the final chart version remain cleared from the resolved inputs.

alex-au:
## Description

The interactive `block node install` / `upgrade` / `reconfigure` wizard rendered the individual storage paths across **three** pages: `archive/live/log` on one page, then one page each for the optional paths that apply to the live chart version (e.g. `plugins`, then `application-state`). Operators expect all the individual storage paths on a **single** page.

Root cause: `huh` v1.0.0 can hide only whole `huh.Group`s, not individual fields, and each optional storage path must appear/disappear reactively as the operator edits the chart version earlier in the same wizard (introduced in #827 / #826). To satisfy that, `AddStoragePathPrompts` placed each optional path in its **own** hide-gated group — and each group is a separate wizard page, producing the multi-page layout.

The fix keeps all individual paths in a **single** `huh.Group` (one page) while preserving the per-field reactivity. Because `huh` renders every field in a group and the built-in `huh.Input` never reports `Skip()==true`, a small `spacedField` wrapper intercepts `Skip()` / `View()`: while its live chart-version predicate says an optional path does not apply, the field is skipped during navigation and rendered empty. `huh` recomputes field positions on every keypress (`Form.UpdateFieldPositions`), so a predicate that flips mid-wizard — an optional path becoming (in)applicable after a chart-version edit — is honored live. The `afterRun` normalization (clearing the unused mode's targets and any non-applicable optional path) is unchanged.

**Even spacing:** `huh` renders its `FieldSeparator` between every adjacent pair of fields, including around a field that renders empty, so a hidden path in the middle of the list (e.g. `verification` at `>=0.37.0`, which sits between `log` and `plugins`) would otherwise leave a doubled blank line. To keep spacing even regardless of which paths are hidden, the storage group runs with `huh`'s built-in separator zeroed (`soloThemeNoFieldSeparator`, re-applied via a per-group theme override because `huh.Form.WithTheme` clobbers per-group themes), and each `spacedField` prepends its own single leading gap only when an earlier field is visible.

### Files changed

| File | Change |
|---|---|
| `internal/ui/prompt/spaced_field.go` | New `spacedField` wrapper around `huh.Field`: `Skip()` / `View()` gated on a live `hidden()` predicate, an owned leading gap via `leadingGap()`, and `Update()` re-wraps the returned model so `huh` keeps addressing the wrapper. |
| `internal/ui/prompt/blocknode.go` | `AddStoragePathPrompts` builds one individual-paths group (all fields wrapped in `spacedField`, optional ones gated on the live chart version) instead of a core group plus one group per optional path; registers the zeroed-separator theme on the group; updated docstring. |
| `internal/ui/prompt/theme.go` | New `soloThemeNoFieldSeparator()` — `SoloTheme` with the inter-field separator removed. |
| `internal/ui/prompt/prompt.go` | `Wizard.overrideTheme()` records a per-group theme re-applied after the form's shared theme in `Run()`. |
| `internal/ui/prompt/spaced_field_test.go` | Unit tests for the wrapper's `Skip()`/`View()` toggling and leading-gap behavior. |
| `internal/ui/prompt/wizard_test.go` | Group-count assertions updated 6 → 3 (mode + base + one individual-paths group). |

### Behavior note

Individual-paths mode now renders on a **single** page (core `archive/live/log` plus every applicable optional path), with even spacing between the shown paths regardless of which optional path is hidden. Full Shift+Tab / Tab back/forward navigation is retained, and optional-storage applicability is unchanged (`verification` `[0.26.2, 0.37.0)`, `plugins` `[0.28.1, ∞)`, `application-state` `[0.37.0, ∞)`).

### Review guide

**Checklist**
- `spaced_field.go` — `Skip()` returns `true` while `hidden()`; `View()` returns `""` while `hidden()` and otherwise prepends `fieldGap` iff `leadingGap()`; `Update()` stores the delegated model back (`h.Field = m.(huh.Field)`) and returns the wrapper, so `huh`'s `g.selector.Set(i, m.(Field))` keeps the override in place. Wrapper depends only on the exported `huh.Field` interface (no concrete `*huh.Input` assertions exist in `huh`).
- `fieldGap` (`"\n\n"`) mirrors `huh`'s default `FieldSeparator`; the group must run with the separator zeroed or gaps double up.
- `blocknode.go` `AddStoragePathPrompts` — all individual fields are wrapped; each `hidden`/`leadingGap` re-reads a per-entry visibility predicate that re-reads `*chartVersion`. The single group keeps `WithHideFunc(!inIndividualMode())` so base-path mode hides the whole page. `storagePathField`'s validator stays suppressed while the field is not visible, so a hidden field never contributes a blocking `Error()`.
- `prompt.go` — `overrideTheme` re-applies after `form.WithTheme(SoloTheme())`; verified `huh.Form.Init` does not re-theme groups, so the override survives to render time.
- `afterRun` clears optional targets that do not apply to the final chart version — unchanged.
- No CLI flag/subcommand changed → no `docs/quickstart.md` update required.

**Tests**
```bash
go test -race -tags='!integration' ./internal/ui/prompt/...
task lint
```

**Manual UAT** — `sudo solo-provisioner block node install` on a TTY (no storage flags):
- Advance to the storage mode page and choose **Individual paths** → forward → a **single** page prompts `archive`, `live`, `log`, and every optional path applicable to the current chart version (e.g. `plugins`, `application-state`). There are **no** separate one-field pages, and the blank line between every shown path is uniform (no doubled gap where `verification` is hidden).
- Shift+Tab back to the chart-version page, change it across an optional-storage boundary (e.g. to a `<0.37.0` value like `0.35.1`), forward again → on the same individual-paths page `verification` now appears and `application-state` disappears, still evenly spaced.
- Change it back to a `0.37.x` value → `application-state` returns and `verification` disappears — still all on one page.
- Complete the wizard → the printed "Selected Inputs" summary lists only the paths applicable to the final chart version; non-applicable optional paths are absent.
- Regression: `--non-interactive` and fully-flagged runs bypass all prompts.

**Risks / rollback** — The change is confined to the prompt layer (no state/schema/workflow/template impact). It depends on `huh` v1.0.0 internals (renders every field's `View()`, calls `Skip()` during navigation, recomputes field positions on each keypress, reads the group theme's `FieldSeparator` at render time); a `huh` upgrade should re-verify these. Rollback = revert to one hide-gated group per optional path.

### Related Issues

* Closes #834


brunodam (approved):
LGTM

PR #838, opened 2026-07-11
fix(blocknode): wire application-state path into reality checker and state reader

### LINKED ISSUE #837: `block node upgrade --with-reset` skips wiping the `application-state` directory when storage paths are not in the active config
## Problem

When \`block node upgrade --with-reset\` is run with a different (or absent) \`--config\` file than the one used during install, the \`application-state\` storage directory is not wiped, leaving stale block-node state on disk.

Root cause is two missing pieces added in the \`application-state\` volume PR but never wired into the read path:

1. **Reality checker** (\`internal/reality/blocknode_checker.go\`): the \`switch\` that maps PV claim names to \`BlockNodeStorage\` fields handles \`live\`, \`archive\`, \`log\`, \`verification\`, and \`plugins\`, but has no case for \`application-state\`. So \`BlockNodeState.Storage.ApplicationStatePath\` is always empty after a reality refresh.

2. **State reader** (\`internal/state/state_reader.go\`): the anonymous YAML struct for \`blockNodeState.storage\` lists \`basePath\`, \`archivePath\`, \`livePath\`, \`logPath\`, \`verificationPath\`, \`pluginsPath\` — but not \`applicationStatePath\`. The field is never round-tripped through the state file.

Because both the reality and state sources return an empty \`ApplicationStatePath\`, the storage resolver falls back to config. If the active config doesn't specify the path (e.g. the operator used \`-c ./custom-config.yaml\` for install but omits it on upgrade), \`ResolveStoragePaths\` derives a relative \`"application-state"\` path, \`ClearStorageDirectory\` finds no such directory, and silently skips it. The other directories (archive, live, logs, plugins) are read correctly from the PVs and are wiped.

## Proposed fix

* Add an \`application-state\` case to the reality checker's PV switch to populate \`ApplicationStatePath\` and \`ApplicationStateSize\`.
* Add \`applicationStatePath\` (and \`applicationStateSize\`) to the state reader's YAML struct and the corresponding struct-to-model mapping.

## Acceptance

- [ ] \`block node upgrade --with-reset\` clears the \`application-state\` directory even when \`-c\` points to a different config than the one used at install time
- [ ] Reality checker correctly reflects \`ApplicationStatePath\` from the installed PV
- [ ] State round-trip preserves \`ApplicationStatePath\` across CLI invocations

brunodam:
## Description

`block node upgrade --with-reset` silently skipped clearing the `application-state`
directory whenever the command was run with a different (or absent) `--config` than
the one used at install time. Every other storage directory (archive, live, logs,
plugins) was wiped correctly.

Two pieces that enumerate block-node storage paths both missed `application-state`
when that volume was introduced:

1. **Reality checker** (`internal/reality/blocknode_checker.go`) — the `switch` that
   maps PV claim names to `BlockNodeStorage` fields had no case for
   `application-state-storage-pvc`, so `ApplicationStatePath` was always empty after
   a cluster state refresh.
2. **State reader** (`internal/state/state_reader.go`) — the YAML struct for
   `blockNodeState.storage` listed every other path field but not
   `applicationStatePath`, so the value was never round-tripped from disk.

With both sources returning `""`, `ResolveStoragePaths` derived the relative path
`"application-state"`, `DirExists` found nothing, and the wipe was silently skipped.

Both fixes are additive only: the new switch case fires exclusively for
`application-state-storage-pvc`, and adding the field to the state YAML struct is
backward-compatible (existing state files without the key unmarshal to `""`, the
same default as before).

### Related Issues

* Closes #837

### DIFF
--- a/internal/reality/blocknode_checker.go
+++ b/internal/reality/blocknode_checker.go
@@ -210,6 +210,9 @@ func (b *blockNodeChecker) populateStorageFromPVs(
 		case strings.Contains(claimName, "plugins"):
 			storage.PluginsPath = hostpath
 			storage.PluginsSize = size
+		case strings.Contains(claimName, "application-state"):
+			storage.ApplicationStatePath = hostpath
+			storage.ApplicationStateSize = size
 		}
 
 	}
--- a/internal/reality/blocknode_checker_test.go
+++ b/internal/reality/blocknode_checker_test.go
@@ -1,0 +1,119 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build !integration
+
+package reality
+
+import (
+	"testing"
+
+	"github.com/hashgraph/solo-weaver/pkg/models"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func makePV(claimNamespace, claimName, size, hostPath string) unstructured.Unstructured {
+	return unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"spec": map[string]interface{}{
+				"claimRef": map[string]interface{}{
+					"namespace": claimNamespace,
+					"name":      claimName,
+				},
+				"capacity": map[string]interface{}{
+					"storage": size,
+				},
+				"hostPath": map[string]interface{}{
+					"path": hostPath,
+				},
+			},
+		},
+	}
+}
+
+func TestPopulateStorageFromPVs_ApplicationStatePVC(t *testing.T) {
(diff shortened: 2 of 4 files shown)


tomzhenghedera (dismissed):

Nana-EC (approved):
LG,
to confirm we're keeping verification around for backwards compatibility but it doesn't break anything if it's not present?

tomzhenghedera (approved):

alex-au (approved):
LGTM!

PR #876, opened 2026-07-15
feat(blocknode): add statusz REST client for BN traffic-shaper monitor

### LINKED ISSUE #751: Story 3.7 — statusz REST client (`inbound-clients`, `outbound-clients`)

Design: [v4 design](https://github.com/automa-saga/traffic-shaper/blob/main/docs/bn-qos-multiclass-priority-design-v4-nft-priority.md) §7.1, §8.1.2, §10 risk 11.

Implement the REST client the poll loop uses to read the BN's statusz endpoints and decode the `network-data` schema.

**Acceptance criteria:**
- GET `statusz/inbound-clients` and `statusz/outbound-clients` every 5 s; decode `NetworkData` / `NetworkConnection` (category, remote address/port, tls_required).
- Transport is **REST/JSON, not gRPC** (the `network-data.proto` is only the data *shape*).
- The contract is **provisional**: build against a **mock REST statusz server** returning `NetworkData`-shaped JSON for daemon dev/tests, and keep decode behind a thin layer so schema evolution is absorbed without churning the poll loop.
- **Blocking dependency:** confirm the schema + sample responses with the BN team before locking the client.
- Surface poll errors to the outage policy (#755) rather than crashing the monitor.

brunodam:
## Description

Implements the REST client the traffic-shaper monitor's poll loop uses to read the Block Node's statusz endpoints and decode the `network-data` payload. Before this, `runStatuszPoll` was a stub and there was no Go representation of the statusz schema; this lands the read+decode half of the poll loop as a standalone, unit-tested unit.

The transport is deliberately plain **REST/JSON over `net/http`, not gRPC** — `network-data.proto` defines only the payload *shape*. Decode is confined to the client's `fetch` boundary and a single set of tagged types, so the still-provisional BN schema can evolve in one place without churning the poll loop or its downstream consumers. Every failure mode returns a wrapped `errorx` error and never panics, so the poll loop can hand faults to the outage policy rather than crashing the monitor.

This PR does **not** wire the client into `runStatuszPoll` (the stub stays), and does not implement category→policy mapping, membership apply, or the 5 s poll cadence / outage-policy behaviour — those are sibling stories in the epic.

### API changes

New in `internal/daemon/blocknode`:

```go
type StatuszClient struct { /* ... */ }
func NewStatuszClient(baseURL string) *StatuszClient
func (c *StatuszClient) InboundClients(ctx context.Context) (NetworkData, error)
func (c *StatuszClient) OutboundClients(ctx context.Context) (NetworkData, error)

type NetworkData struct { ActiveEndpoints []NetworkConnection }
type NetworkConnection struct { Local, Remote Endpoint; Category string; TLSRequired bool }
type Endpoint struct { Address, Port string }
```

`Endpoint.Port` is a string because the BN reports `"*"` (any port) alongside numeric ports, and `Address` may be a CIDR (e.g. `10.10.1.0/24`).

### Review guide

**Schema provenance / provisional contract.** The tagged types mirror the `network-data.proto` message and JSON samples in the traffic-shaper v4 design doc (§7.1, §7.2). The contract is still provisional — the BN team should confirm the live statusz payload matches before this client is considered locked. The tagged types + `fetch` decode in `statusz_client.go` are the single place a schema revision needs to touch; `json.Decode` already ignores the proto fields the traffic-shaper doesn't consume (`scheme`, `protocol`, `certificate`).

**Checklist:**
- `internal/daemon/blocknode/statusz_client.go` — error namespaces: transport/non-200 → `ExternalError`, malformed body → `IllegalFormat`, URL build → `IllegalArgument`. Confirm no path can panic.
- Per-request timeout (`statuszClientTimeout = 4s`) is intentionally below the 5 s poll cadence so a hung response fails fast.
- Category is kept as the raw BN string here — mapping to policy names / nft sets is a separate story.

**Test commands:**

```
go test ./internal/daemon/blocknode/...
task lint
```

`statusz_client_test.go` runs against an `httptest` mock statusz server and covers: inbound/outbound decode against the design doc's verbatim sample payloads, empty endpoints, non-200 → `ExternalError`, malformed JSON → `IllegalFormat`, transport failure, and cancelled context (no panic).

**Risks / rollback:** the client is unreferenced by production paths (not wired into the poll loop), so reverting is a clean file removal with no behavioural impact.

### Related Issues

* Closes #751

### DIFF
--- a/internal/daemon/blocknode/statusz_client.go
+++ b/internal/daemon/blocknode/statusz_client.go
@@ -1,0 +1,119 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package blocknode
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/joomcode/errorx"
+)
+
+// statuszClientTimeout bounds a single statusz request. It is deliberately
+// shorter than the poll loop's 5 s steady-state cadence so a hung BN response
+// fails fast and surfaces to the outage policy rather than stacking requests.
+const statuszClientTimeout = 4 * time.Second
+
+// statusz endpoint paths, relative to the client's base URL. The transport is
+// REST/JSON: network-data.proto defines only the payload shape, not a gRPC
+// service.
+const (
+	inboundClientsPath  = "statusz/inbound-clients"
+	outboundClientsPath = "statusz/outbound-clients"
+)
+
+// The types below mirror the JSON encoding of
+// block-node/api/network-data.proto. They are the decode boundary: the poll
+// loop consumes these types via the client, never raw JSON, so coupling to the
+// (provisional) BN schema stays in this one file. json.Decode drops the proto
+// fields the traffic-shaper doesn't consume (scheme, protocol, certificate), so
+// they are simply absent here rather than modelled and thrown away.
+
+// Endpoint is one side (local or remote) of a NetworkConnection. Port is a
+// string because the BN reports "*" (any port) alongside numeric ports, and
+// Address may be a single IP or a CIDR (e.g. "10.10.1.0/24").
+type Endpoint struct {
+	Address string `json:"address"`
+	Port    string `json:"port"`
+}
+
+// NetworkConnection is one active endpoint reported by a statusz endpoint.
+// Category is left as the raw BN string here — mapping categories to policy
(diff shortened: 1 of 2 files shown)


alex-au (approved):
Nice and clean!

Issue #754, opened 2026-06-25
Story 3.9 — Apply membership deltas via `network policy set --cidrs --name`


alex-au:
**Epic:** #736 — BN traffic-shaper monitor

Design: [v4 design](https://github.com/automa-saga/traffic-shaper/blob/main/docs/bn-qos-multiclass-priority-design-v4-nft-priority.md) §8.1.2, §8.4.5, §8.5, §7.5.

Apply the computed deltas through the `network policy` element verbs so the CLI and the monitor share one code path.

**Acceptance criteria:**
- Apply per-policy deltas via `network policy set --cidrs ... --name ...`, batched into one `nft -f` transaction per policy for atomicity.
- The monitor never calls `policy create` (structure is fixed by `block node install`); it uses only element verbs.
- Serialize the apply with the shared `flock` (`/run/solo-provisioner/network/.applying`, `LOCK_EX|LOCK_NB`): the daemon checks it non-blocking before exec'ing and skips the tick if a prior operator apply is still running, so daemon and hand-run CLI writes to the same `inet weaver` sets cannot interleave (CLI side: #759).
- Both the CLI and monitor paths produce identical kernel state (§8.5).

PR #878, opened 2026-07-15
feat(traffic-shaper): map statusz categories to policies and diff vs live nft sets

### LINKED ISSUE #752: Story 3.8 — Map statusz categories to policy names and diff vs live nft sets

Design: [v4 design](https://github.com/automa-saga/traffic-shaper/blob/main/docs/bn-qos-multiclass-priority-design-v4-nft-priority.md) §8.1.2, §7.5.

Implement the monitor's internal statusz-category→policy-name mapping and the diff against live nft set membership. (Absorbs the former standalone diff-engine story.)

**Acceptance criteria:**
- Internal, non-configurable mapping: inbound `publisher`→`bn-publisher`, `partner`→`bn-partner`, `restricted`→`bn-restricted`; outbound `peer_bn`→`bn-backfill` (compound `ip:port` entries).
- Compute per-policy add/delete deltas by diffing the desired membership against the **live** `nft list table inet weaver` sets — read back from the kernel, not a remembered hash (stateless across daemon restarts, self-correcting against drift).
- Canonicalize both desired and live CIDR lists into a stable **numeric** order (so `10.1.0.2/32` precedes `10.1.0.10/32`) before comparison, and feed the same normalization into the apply args so compare and apply never disagree.
- Category semantics: an **empty** `active_endpoints` list for a category clears its set; a category **absent entirely** from the response leaves its set untouched.
- Deltas reclassify future packets only; in-flight flows keep their class until close (§7.6).
- mgmt sets are operator-curated; the monitor does not mutate them.

brunodam:
## Description

The traffic-shaper monitor keeps the block node's nftables `inet weaver` policy
sets in sync with the categories the BN reports over statusz. This PR adds the
pure middle stage of that reconciliation: the internal category→policy mapping
and the diff that compares desired membership against the **live** kernel sets to
produce per-policy add/delete deltas. It deliberately owns neither the statusz
fetch (#751) nor the delta apply (#754) — those wrap this deterministic core.

The diff reads live membership back from the kernel every time rather than
trusting a remembered hash, so it is stateless across daemon restarts and
self-correcting against drift. Both the desired list and the live list are run
through one canonicalizer before comparison, and the same normalized form is what
the apply path consumes, so compare and apply can never disagree on spelling or
order. A single-host `/32` is collapsed to the bare address nft prints for an
interval `ipv4_addr` set (so `10.1.0.2/32` and `10.1.0.2` are the same element),
and elements sort numerically (`10.1.0.2` before `10.1.0.10`). The compound
`ip:port` conversion for the outbound `bn-backfill` set is now a single exported
helper shared with the `network policy set` apply path, so both render `ip . port`
keys byte-for-byte identically.

Category semantics follow the acceptance criteria precisely: a category present
with an empty endpoint list clears its set; a category absent entirely from the
response leaves its set untouched (and its live set is never even read);
categories with no mapping — the operator-curated mgmt sets — are never touched.

The full poll loop is not wired yet, but `runStatuszPoll` now drives a
not-yet-live `reconcilePolicies` seam (with an empty view, so zero deltas and no
kernel writes) so #751/#754 can plug in fetch and apply without further plumbing.

### Files changed

| File | Change |
|---|---|
| `internal/network/policy/diff.go` | New. `CompoundElement` (exported, shared conversion), `CanonicalizeElements` (numeric sort, `/32` collapse, dedup, unparseable-preserving), `DiffElements` → `SetDelta{Adds,Deletes}`. |
| `internal/network/policy/policy.go` | `setElements` now calls the shared `CompoundElement` instead of inlining the `ip . port` conversion. |
| `internal/daemon/blocknode/statusz_policy.go` | New. Fixed category→policy map, `CategoryEndpoints` input type (present-empty vs absent), `computePolicyDeltas` (build desired, read live via the `Runner` seam, diff). |
| `internal/daemon/blocknode/traffic_shaper_monitor.go` | Monitor holds an `elementLister` (defaults to the policy exec Runner); `runStatuszPoll` drives the new `reconcilePolicies` seam. |
| `internal/network/policy/diff_test.go` | New. Canonicalization ordering / `/32` collapse / dedup / compound / idempotency; diff add/delete/no-op/compound cases. |
| `internal/daemon/blocknode/statusz_policy_test.go` | New. Mapping, empty-clears vs absent-untouched, unmapped-ignored, compound conversion, live-read error propagation, nil-lister guard. |

### Review guide

- **Live read is per-set, not `list table`.** `computePolicyDeltas` reads each
  policy's membership with the existing `Runner.ListElements` (`nft list set inet
  weaver <set>`), which is equivalent to parsing the whole `list table` document
  but reuses the already-tested `parseElementsLine`. Worth a look:
  `internal/daemon/blocknode/statusz_policy.go` `computePolicyDeltas`.
- **Canonicalization matches nft's interval-set rendering.** The `/32`→bare
  collapse in `diff.go` `parseElement` is what makes desired vs live agree; if
  nft rendered single hosts differently this would churn. These sets are
  daemon-owned (sole writer), so the round-trip is self-consistent.
- **Present-empty vs absent** is modeled by map-key presence in
  `CategoryEndpoints`, not an empty slice — verify the two paths in
  `computePolicyDeltas` and their tests.

Test commands:

```bash
go test -tags='!integration' ./internal/network/policy/... ./internal/daemon/blocknode/...
task lint:check
```

No CLI-flag or template changes, so no `docs/quickstart.md` update. Rollback is
dropping the two new files; nothing changes observable behavior until #754 wires
apply into the poll loop.

### Manual UAT

There is no operator-facing runtime path in this PR (the poll loop is stubbed and
writes nothing), so the acceptance criteria are covered by unit tests. The one
thing unit tests can't prove is that the canonical element spelling this PR
produces matches what **real nft** prints back — the assumption the diff's
correctness rests on. Verify that round-trip on a Linux host with `nft` (e.g. the
UTM VM), using throwaway policies so no real BN state is touched:

```bash
# Plain interval ipv4_addr set: a /32 must read back as a bare host, numerically ordered.
sudo solo-provisioner network policy create --name uat-plain \
  --stamp publisher --ports 40840 --pod-cidr 10.4.0.0/24
sudo solo-provisioner network policy set --name uat-plain --cidrs 10.1.0.10/32,10.1.0.2/32
sudo nft list set inet weaver uat-plain
#   expect: elements = { 10.1.0.2, 10.1.0.10 }   ← /32 collapsed, .2 before .10

# Compound ipv4_addr . inet_service set: an ip:port must read back as "<ip> . <port>".
sudo solo-provisioner network policy create --name uat-compound \
  --stamp reserve-egress --reply-stamp backfill-response --pod-cidr 10.4.0.0/24
sudo solo-provisioner network policy set --name uat-compound --cidrs 10.30.5.7:43473
sudo nft list set inet weaver uat-compound
#   expect: elements = { 10.30.5.7 . 43473 }

# cleanup
sudo solo-provisioner network policy delete --name uat-plain
sudo solo-provisioner network policy delete --name uat-compound
```

Both readbacks must match, byte for byte, what `CanonicalizeElements` produces for
the same inputs. If nft spells either form differently, the diff would churn and
`parseElement`'s canonicalization needs to follow nft's rendering.

### Related Issues

* Closes #752

### DIFF
--- a/internal/daemon/blocknode/statusz_policy.go
+++ b/internal/daemon/blocknode/statusz_policy.go
@@ -1,0 +1,140 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package blocknode
+
+import (
+	"context"
+	"sort"
+
+	"github.com/hashgraph/solo-weaver/internal/network/policy"
+	"github.com/joomcode/errorx"
+)
+
+// Category is a statusz traffic category as reported by the block node's statusz
+// endpoints (inbound-clients / outbound-clients). The vocabulary is fixed by the
+// BN's contract, not operator-configurable.
+type Category string
+
+const (
+	// CategoryPublisher is the inbound publisher category.
+	CategoryPublisher Category = "publisher"
+	// CategoryPartner is the inbound partner category.
+	CategoryPartner Category = "partner"
+	// CategoryRestricted is the inbound restricted category.
+	CategoryRestricted Category = "restricted"
+	// CategoryPeerBN is the outbound peer-block-node category. Its endpoints are
+	// compound ip:port pairs — the backfill set is keyed on destination address
+	// AND port.
+	CategoryPeerBN Category = "peer_bn"
+)
+
+// categoryBinding is the fixed mapping from a statusz category to the nft policy
+// set it drives and how that set is keyed.
+type categoryBinding struct {
+	// policyName is the nft set (network policy) the category's endpoints
+	// populate.
+	policyName string
+	// compound is true when the set is a compound `ipv4_addr . inet_service`
+	// key set fed ip:port endpoints (bn-backfill); false for plain ipv4_addr
+	// sets fed host/CIDR endpoints.
+	compound bool
+}
+
+// categoryBindings is the internal, non-configurable statusz-category → policy
+// mapping. Both the BN's category vocabulary and the provisioner's policy names
+// are fixed in code, so this is a static table rather than config. Categories
+// not listed here — including the operator-curated mgmt sets — are never touched
+// by the monitor.
+var categoryBindings = map[Category]categoryBinding{
+	CategoryPublisher:  {policyName: "bn-publisher"},
+	CategoryPartner:    {policyName: "bn-partner"},
+	CategoryRestricted: {policyName: "bn-restricted"},
+	CategoryPeerBN:     {policyName: "bn-backfill", compound: true},
+}
+
+// CategoryEndpoints is one poll's desired membership view, keyed by statusz
+// category. For each category PRESENT in the map, the slice is its active
+// endpoints: plain IPv4 hosts/CIDRs for inbound categories, "ip:port" pairs for
+// the outbound peer_bn category. The present-vs-absent distinction is load
+// bearing and preserved end to end:
+//
+//   - a category present with an empty (or nil) slice clears that policy's set;
+//   - a category absent from the map leaves that policy's set untouched.
+//
+// This is the minimal shape the diff needs; the statusz client (#751) decodes the
+// wire NetworkData response into it, and the apply path (#754) consumes the
+// resulting deltas.
+type CategoryEndpoints map[Category][]string
+
+// PolicyDelta is the computed membership change for one policy set: the policy
+// (nft set) name and the canonicalized add/delete lists.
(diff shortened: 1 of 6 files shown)


alex-au (approved):
LGTM!

PR #886, opened 2026-07-16
feat(daemon): delegate privileged ops via sudo solo-provisioner exec

### LINKED ISSUE #774: Story 3.6 — Privileged-exec delegation (unprivileged daemon → `sudo solo-provisioner`)

Design: [v4 design](https://github.com/automa-saga/traffic-shaper/blob/main/docs/bn-qos-multiclass-priority-design-v4-nft-priority.md) §8.1, §8.5.

Implement the privilege model: the daemon runs unprivileged and delegates every privileged operation by exec'ing the `solo-provisioner` binary under sudo.

**Acceptance criteria:**
- The daemon runs as `User=weaver`; privileged ops (nft `add element`, `tc qdisc add` on a new veth) are delegated by exec'ing `sudo solo-provisioner …`.
- A sudoers grant restricts the privilege to the `solo-provisioner` binary only.
- Covers both the poll loop's `network policy set` and the watcher's `block node tc-attach`.

brunodam:
## Description

The `solo-provisioner-daemon` runs unprivileged (systemd `User=weaver`) but the
traffic-shaper monitor's work requires root: writing nftables set membership and
attaching tc qdiscs to per-pod veths. Rather than give the daemon capabilities or
run it as root, it delegates each privileged action by exec'ing the root
`solo-provisioner` CLI under `sudo`. This PR implements that privilege mechanism.

`internal/daemon/privexec` is a small `Delegator` that runs
`sudo solo-provisioner <args…>`: it resolves both binaries by absolute path (the
CLI sibling of the running daemon first, then the two sudoers-granted install
paths — never a bare `PATH` lookup), runs synchronously, returns stdout, and maps
every failure — missing `sudo`, missing CLI, denied grant, non-zero exit — to a
`*daemonkit.ProbeError` carrying an operator-facing reason and resolution so the
consumer can surface it through `/status`. A typed `NetworkPolicySet(name, cidrs)`
wraps the poll loop's `network policy set` write path; the generic `Run` seam is
what the pod watcher's `block node tc-attach` will use.

This is intentionally the **mechanism only**. Both call sites live in
not-yet-landed stories (the poll-loop apply and bootstrap/outage policy; the
tc-attach verb), so the monitor merely *holds* the delegator (defaulted in its
constructor, like the existing live-set reader) — those stories consume it with no
constructor change. Two of the story's acceptance criteria — the `User=weaver`
service unit and the whole-binary sudoers grant — already shipped with the
self-upgrade epic; this PR locks them with a regression test rather than
re-authoring them.

`privexec` is the only package in the daemon's import closure that reaches for
`os/exec`, and it execs nothing but `sudo` + `solo-provisioner`, so the "no raw
tools" guarantee in the security model still holds even though `os/exec` is now
present. The daemon architecture doc is updated to say exactly that.

### Files changed

| File | Change |
|---|---|
| `internal/daemon/privexec/delegate.go` | New. `Delegator` interface + `execDelegator`: absolute-path resolution of `sudo` and `solo-provisioner`, `sudo <cli> <args…>` via `exec.CommandContext(...).Output()`, failure → `*daemonkit.ProbeError` (reason + resolution), typed `NetworkPolicySet`. |
| `internal/daemon/privexec/delegate_test.go` | New. 9 macOS-safe tests via injected exec/stat/executable seams: argv shape, resolution order + fallback, empty-cidrs omits `--cidrs`, empty-name / no-arg guards, real `*exec.ExitError` → `ProbeError` (stderr surfaced, underlying error still unwrappable). |
| `internal/daemon/blocknode/traffic_shaper_monitor.go` | `TrafficShaperMonitor` holds a `privexec.Delegator`, defaulted to `privexec.New()` in the constructor. No behavioural change to the stubs. |
| `internal/templates/weaver_privilege_test.go` | New. Locks the two supporting invariants: sudoers grants only the `solo-provisioner` binary (no raw `kubectl`/`helm`/`kubeadm`/`systemctl`/`nft`/`tc`/`mkdir`); the daemon service unit keeps `User=weaver`/`Group=weaver` with no `NoNewPrivileges=true`. Comment prose is stripped so assertions match the effective directives. |
| `docs/dev/daemon/daemon-architecture.md` | Security posture: `os/exec` is now present, scoped to `internal/daemon/privexec`; "no raw tools" guarantee retained; CLI-binary hash/signature verification noted as still pending. |

### Review guide

- **Binary resolution never trusts `PATH`.** `resolveCLI` prefers the CLI
  installed alongside the daemon (`os.Executable()` dir) and falls back to the two
  paths the sudoers file grants (`/opt/solo/weaver/bin`, `/usr/local/bin`); `sudo`
  is resolved from an absolute candidate list the same way the nft/tc runners do.
  Worth confirming the fallback order in `delegate.go` `resolveCLI` /
  `resolve`.
- **stdout stays clean.** `Run` returns stdout only; stderr is folded into the
  `ProbeError` message (via `*exec.ExitError.Stderr`), so a future caller can
  parse `--output json` from stdout without stderr contamination.
- **Failure surfacing.** Every error path returns a `*daemonkit.ProbeError` with a
  stable `Reason` and an actionable `Resolution` (naming the sudoers file and the
  exact reproduce command) — the same reason/resolution shape the upgrade monitor
  feeds into `/status`.
- **Invariant test reads directives, not comments.** Both the sudoers file and the
  service unit deliberately *name* the things they exclude (raw tools;
  `NoNewPrivileges`), so `weaver_privilege_test.go` strips `#` lines before
  asserting — see `directiveLines`.

Test commands:

```bash
go test -tags='!integration' ./internal/daemon/privexec/... ./internal/daemon/blocknode/... ./internal/templates/...
task lint:check
```

No CLI-flag or embedded-template changes, so no `docs/quickstart.md` update and no
upgrade-path concern. Rollback is dropping the new package and reverting the two
one-line edits; nothing changes observable behavior until a later story wires a
call site.

### Manual UAT

The delegator has no live caller yet, so end-to-end shaping isn't observable in
this PR. What *is* verifiable is the escalation path it depends on — that the
unprivileged `weaver` user can exec `sudo solo-provisioner` non-interactively via
the installed sudoers grant. On a provisioned host (or UTM VM):

```bash
# As the weaver user, the whole-binary NOPASSWD grant must let a subcommand run
# as root with no password prompt — this is exactly what privexec.Run performs.
sudo -n -u weaver sudo -n /opt/solo/weaver/bin/solo-provisioner version
#   expect: version output, no password prompt, exit 0

# A denied/absent grant surfaces as the ProbeError resolution the daemon logs:
#   "verify the weaver sudoers grant (/etc/sudoers.d/solo-provisioner) permits
#    `sudo solo-provisioner` …"
sudo cat /etc/sudoers.d/solo-provisioner
#   expect: grants only /opt/solo/weaver/bin/solo-provisioner and
#           /usr/local/bin/solo-provisioner (bare + " *"), no raw tools
```

### Related Issues

* Closes #774

### DIFF
--- a/docs/dev/daemon/daemon-architecture.md
+++ b/docs/dev/daemon/daemon-architecture.md
@@ -166,10 +166,18 @@ systemd unit), never as root. It holds no Linux capabilities of its own.
 
 **Privilege via exec-delegation, not raw tools.** When the daemon needs to perform a
 privileged action it execs the root `solo-provisioner` CLI (see "CLI invocation & the import
-boundary" above) — it does **not** invoke `kubectl`/`helm`/`kubeadm`/`systemctl` directly.
-This is verified, not aspirational: the daemon binary imports `os/exec` nowhere in its
-transitive closure, talks to Kubernetes via client-go, and talks to systemd via the
-go-systemd dbus API. Its only privileged-escalation path is the sudoers grant below.
+boundary" above) — it does **not** invoke `kubectl`/`helm`/`kubeadm`/`systemctl`/`nft`/`tc`
+directly. It talks to Kubernetes via client-go and to systemd via the go-systemd dbus API;
+its only privileged-escalation path is the sudoers grant below. The exec-delegation itself is
+confined to a single package, `internal/daemon/privexec`, which is the **only** place in the
+daemon's import closure that reaches for `os/exec`. It execs nothing but `sudo` +
+`solo-provisioner`, so the "no raw tools" guarantee holds even though `os/exec` is now present.
+The delegation is synchronous (run, wait, capture stdout) and execs `sudo -n` so a
+missing or expired grant fails fast instead of blocking on a prompt the daemon has no
+tty for. Any failure is returned as a `*daemonkit.ProbeError` carrying an operator-facing
+reason/resolution, which the daemon boundary converts into a `daemonkit.StatusError` so
+`/status` explains it. The self-upgrade protocol's detached child spawn is a separate exec
+mechanism with different lifecycle semantics.
 
 **Sudoers grant (`internal/templates/files/weaver/sudoers`).** Weaver is granted passwordless
 sudo to the `solo-provisioner` binary as a whole (any subcommand), at both its install paths:
--- a/internal/daemon/blocknode/traffic_shaper_monitor.go
+++ b/internal/daemon/blocknode/traffic_shaper_monitor.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/automa-saga/logx"
+	"github.com/hashgraph/solo-weaver/internal/daemon/privexec"
 	"github.com/hashgraph/solo-weaver/internal/network/policy"
 	"github.com/joomcode/errorx"
 )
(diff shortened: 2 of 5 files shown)


alex-au (approved):
LGTM, mechanism only, well tested!

PR #889, opened 2026-07-16
fix: skipped migrations due to absent state yaml 

### LINKED ISSUE #789: Absent `state.yaml` causes all `CLIVersionMigration` startup migrations to silently no-op on existing clusters
## Problem

UAT testing of PR #788 (cilium host-legacy-routing migration) revealed that the migration never triggers on existing clusters. The root cause is that `state.yaml` is not created during CLI install, so `state.provisioner.version` is absent on disk.

`RunStartupMigrations` calls `state.ReadProvisionerVersionFromDisk()` to populate `installedCLIVersion` in the migration context. When `state.yaml` does not exist the function returns `""`. `CLIVersionMigration.Applies()` treats an empty installed version as a fresh install and immediately returns `false`:

```go
// internal/migration/cli_version_migration.go
if installedCLIVersion == "" {
    return false, nil // fresh install — no previous version to migrate from
}
```

On a host that already has a provisioned cluster but no `state.yaml` (because the file was never written, or was written to a different path by an earlier CLI version), every `CLIVersionMigration` startup migration silently no-ops. This affects all three currently registered startup migrations:

- `CiliumAccelerationMigration` (boundary: v0.19.1)
- `CiliumAgentRestartMigration` (boundary: v0.20.0)
- `CiliumHostLegacyRoutingMigration` (boundary: v0.21.0, PR #788)

None of them ever apply on an existing cluster that lacks a `state.yaml`, regardless of how far behind the cluster is.

## Proposed fix

In `RunStartupMigrations` (`cmd/cli/commands/common/run.go`), when `ReadProvisionerVersionFromDisk` returns an empty string, default `installedCLIVersion` to `"0.0.0"` rather than leaving it empty. This makes every version-boundary migration apply on first run, which is correct for any cluster that pre-dates state tracking.

The individual migrations are already self-guarding: each `Execute()` calls `kubernetesInstalled()` and skips if Kubernetes is not present. So defaulting to `"0.0.0"` on a genuinely fresh install (no k8s, no state file) is safe — `Applies()` would return `true` but `Execute()` would skip, leaving behavior unchanged.

## Acceptance

- [ ] On a host with an existing cluster and no `state.yaml`, running the provisioner applies all pending `CLIVersionMigration` startup migrations (e.g. `enable-host-legacy-routing` is set via `cilium upgrade`)
- [ ] On a host with no cluster and no `state.yaml` (fresh install), all startup migrations skip cleanly without error
- [ ] On a host where `state.yaml` exists with a real version, behaviour is unchanged

### Related Issues

* Closes #781

### LINKED ISSUE #790: Write `state.yaml` with current CLI version at the end of `kube cluster install`
## Problem

`state.yaml` is not written during `kube cluster install`, so a freshly provisioned cluster has no on-disk provisioner version. On the next CLI upgrade, `ReadProvisionerVersionFromDisk()` returns `""` and all `CLIVersionMigration` startup migrations silently no-op (see #789).

The `"0.0.0"` fallback proposed in #789 handles clusters already in the field, but it is a heuristic. The correct long-term fix is to make the install workflow write `state.yaml` with the current CLI version so that future upgrades have a real version to compare against — no fallback needed.

## Proposed fix

At the end of the `kube cluster install` workflow, ensure `FlushState()` (or `FlushAll()`) is called with the state manager already holding `ProvisionerState.Version = version.Get().Version`. This is what `Refresh()` stamps today (`state_manager.go:208`), so the version field is correct as long as the flush actually happens.

`solo-provisioner install` (CLI self-install) is **not** the right place — the cluster does not exist yet and the state directory may not be set up. `block node install` runs after `kube cluster install`, so fixing the latter is sufficient.

## Acceptance

- [ ] After `kube cluster install` completes, `state.yaml` exists and `state.provisioner.version` matches the running CLI version
- [ ] A subsequent CLI upgrade correctly detects the installed version and applies any pending `CLIVersionMigration` startup migrations without relying on the `"0.0.0"` fallback from #789

### Related Issues

* Follows from #789

radkomih:
## Description

  - Treat an absent state.yaml (empty on-disk CLI version) as the 0.0.0 baseline in RunStartupMigrations, so pending version-boundary startup migrations run on existing clusters instead of being silently skipped as
  a fresh install.
  - Persist the running CLI version to state.yaml after the startup-migration pass — gated on a version change and a provisioned host (KubernetesInstalled()), and recorded regardless of whether a migration applied
  — so boundary migrations (notably the non-idempotent Cilium agent restart) don't re-run on every subsequent invocation.
  - Add a RecordProvisionerVersion step at the end of kube cluster install so freshly provisioned clusters write a state.yaml with the current version, instead of relying on the 0.0.0 fallback.
  - Add state.PersistProvisionerVersion (Refresh-then-flush, preserving existing state) and models.SetPaths test helper; keep version recording best-effort (warn, never fail startup).

###  Related Issues

  - Closes #789
  - Closes #790

---

## UAT - fix absent state.yaml silent migration no-op

Manual test scenarios, runs in the UTM VM.

### Setup 

```bash
# host (macOS)
# VERSION defaults to 0.0.0, which is below every migration boundary (0.19.1 / 0.19.2 / 0.21.0),
# so an unstamped binary can NEVER show migrations running. Stamp above the highest boundary.
task build:cli GOOS=linux GOARCH=arm64 VERSION=0.99.0
task vm:sync
task vm:ssh
```

```bash
# inside VM: install to expected path so global checks pass
sudo /mnt/solo-weaver/bin/solo-provisioner-linux-arm64 install
BIN=/opt/solo/weaver/bin/solo-provisioner
STATE=/opt/solo/weaver/state/state.yaml
LOG=/opt/solo/weaver/logs/solo-provisioner.log
$BIN version
```

Trigger used by all scenarios (runs migration pass, then prints help):
```bash
sudo truncate -s 0 $LOG && sudo $BIN --log-level debug >/dev/null 2>&1
```

Inspect the migration:
```bash
sudo grep -aE "Running startup migrations|No startup migrations needed|Kubernetes not installed|Kubernetes API not reachable|Cilium not installed|already disabled|Failed to record provisioner version" $LOG
```

---

### Scenario A - absent state.yaml on provisioned host

```bash
sudo mkdir -p /etc/kubernetes && sudo touch /etc/kubernetes/admin.conf
sudo rm -f $STATE
sudo truncate -s 0 $LOG && sudo $BIN --log-level debug >/dev/null 2>&1
sudo grep -A1 provisioner $STATE
```

**Expect:** log shows `Running startup migrations...` (evaluated against 0.0.0 baseline; cilium
migrations then self-skip: `Kubernetes API not reachable; skipping`), and `state.yaml` now exists
with `version: <$CUR>`

---

### Scenario B - fresh machine, no cluster (must not create state)

```bash
sudo rm -f /etc/kubernetes/admin.conf
sudo rm -f $STATE
sudo truncate -s 0 $LOG && sudo $BIN --log-level debug >/dev/null 2>&1
ls -l $STATE  # expect: No such file or directory
```

**Expect:** no `state.yaml` written - `KubernetesInstalled()` gate skips persist.

---

### Scenario C - version already current (idempotent)

```bash
sudo touch /etc/kubernetes/admin.conf
sudo stat -c '%Y %s' $STATE
sudo truncate -s 0 $LOG && sudo $BIN --log-level debug >/dev/null 2>&1
sudo stat -c '%Y %s' $STATE   # expect unchanged
```

**Expect:** persist skipped (onDisk == current), file untouched.

---

### Scenario E - persist failure is best-effort

```bash
sudo touch /etc/kubernetes/admin.conf
sudo rm -f $STATE
sudo chattr +i /opt/solo/weaver/state 2>/dev/null || sudo chmod 500 /opt/solo/weaver/state
sudo $BIN --log-level debug; echo "exit=$?"
# restore
sudo chattr -i /opt/solo/weaver/state 2>/dev/null || sudo chmod 755 /opt/solo/weaver/state
```

**Expect:** `exit=0` + warning `Failed to record provisioner version...` command does not abort.

---

### Scenario D - `kube cluster install` records version (real cluster)

```bash
# host
task vm:reset    
task build:cli GOOS=linux GOARCH=arm64 VERSION=0.99.0
task vm:sync
task vm:ssh
```

```bash
# inside VM
sudo /mnt/solo-weaver/bin/solo-provisioner-linux-arm64 install
sudo /opt/solo/weaver/bin/solo-provisioner kube cluster install  
sudo grep -A1 provisioner /opt/solo/weaver/state/state.yaml
```

**Expect:** `state.yaml` with version, install TUI shows step
"Recording provisioner version" -> "Provisioner version recorded". Only scenario exercising
`steps.RecordProvisionerVersion()` (cluster.go) for real


internal/workflows/migration_cilium_acceleration.go
```diff
@@ -142,6 +142,13 @@ func (m *CiliumAccelerationMigration) Rollback(ctx context.Context, mctx *migrat
 	return nil
 }
 
+// KubernetesInstalled reports whether Kubernetes is provisioned on this host,
+// using the same probe the Cilium migrations gate Execute() on. The startup
+// orchestrator uses it to decide whether to record the provisioner version, so a
+// fresh machine keeps having no state file. A var so tests can stub the probe
+// (which reads /etc/kubernetes/admin.conf, absent on a CI host).
+var KubernetesInstalled = func() bool { return kubernetesInstalled() }
```
alex-au: `KubernetesInstalled` wraps `kubernetesInstalled` seems a bit confusing and the Cilium package tests stub `kubernetesInstalled` may affect `KubernetesInstalled` unintendedly. 
Should we point `KubernetesInstalled` at `defaultKubernetesInstalled` directly to match how the existing `kubernetesInstalled` var is initialized so the two are independent?

pkg/models/paths.go
```diff
@@ -27,3 +27,16 @@ var (
...
 }
+
+// SetPaths re-roots the process-wide Weaver paths at the given home directory and
+// returns a function that restores the previous paths. It exists so tests can
+// redirect path lookups that are hard-wired to Paths() (e.g. the state-file
+// readers in internal/state) at a temporary directory. Not for production use.
+//
+//	restore := models.SetPaths(t.TempDir())
+//	defer restore()
+func SetPaths(home string) func() {
```
alex-au, line 38: Just a nit, this function is only used in tests, so ideally it lives in the test package, but I see it's more convenient to have it here. Just see if you have a better place for it, otherwise I think it's fine

alex-au (changes requested):
Thank you for the PR, just one nitpick about`KubernetesInstalled` in the comment below if you can take a look.
It would be great if you can add a section of UAT test cases in the PR description and manually test them ([an example](https://github.com/hashgraph/solo-weaver/pull/896)).

pkg/models/paths.go
radkomih: I will leave it for now, because I would have to introduce more changes

alex-au (approved):
Nice work, thank you!

Issue #738, opened 2026-06-25
Epic TS_1 — solo-provisioner network command family


alex-au:
Design: [v4 design](https://github.com/automa-saga/traffic-shaper/blob/main/docs/bn-qos-multiclass-priority-design-v4-nft-priority.md) §8.4.

Implement the operator-facing `solo-provisioner network` command family — the single, generic, command-driven surface for all nft + tc + MetalLB state behind the BN traffic shaper (v4). Four scopes:

- `network firewall` — node-level allowlist (`inet host`).
- `network policy` — per-category nft rules + priority class (`inet weaver`), with `--stamp`/`--deny`/`--reply-stamp`, a policy registry, and tier-order chain re-render.
- `network shape` — tc HTB classes and device roots (the bandwidth plane on `$VETH` and `$EGRESS`).
- `network lb` — MetalLB pools/advertisements.

Every mutation applies to the live kernel **and** atomically rewrites its on-disk artifact under `/etc/solo-provisioner/` (replayed at boot by TS_2's oneshot units). The static plane is rendered exclusively through these commands; `block node install` (TS_2) and the daemon monitor (TS_3) invoke them. This is the foundation epic.

**Note (design decision):** direction and device are derived, not independently-settable flags. Every class in the design's §5 mark map has exactly one direction, so `network policy create` derives `--direction` from `--stamp`, and `network shape create` derives `--device` from `--class`; neither scope takes a free-standing direction/device flag on those forms (see design doc §5/§8.4.2/§8.4.3).

brunodam:
**CLI-surface note — `network firewall` ICMP is a fixed ruleset (no flags)**

Landed in #793 (Story 1.1, #757): `network firewall create` **no longer exposes `--icmp-mgmt` / `--icmp-public`**. ICMP is now a fixed, safe ruleset rather than an operator toggle:

- Full ICMP from the `@mgmt_addrs` allowlist (`echo-request`/`echo-reply`, `destination-unreachable`, `time-exceeded`, `parameter-problem`).
- From every other source: `destination-unreachable` (Path MTU Discovery) and `time-exceeded` (traceroute) are **always** accepted, and `echo-request` is rate-limited to `10/second`.

[...]

PR #891, opened 2026-07-17
feat(network): reject overlapping specific policies and tiebreak tier order by created_at (#772)

### LINKED ISSUE #772: Story 1.5 — Policy registry and tier-order chain re-render
**Epic:** #738 — solo-provisioner network command family

Design: [v4 design](https://github.com/automa-saga/traffic-shaper/blob/main/docs/bn-qos-multiclass-priority-design-v4-nft-priority.md) §8.4.7.

Implement the policy registry (`/etc/solo-provisioner/policies/<name>.json`) and the deterministic tier-order re-render of the full `inet weaver` chain on every `policy create`/`delete`.

**Acceptance criteria:**
- Each `policy create`/`delete` writes/removes the registry JSON (name, action, stamp, reply_stamp, direction, ports, from_entity_world, created_at) and re-renders `network-weaver.nft` from the registry, applied atomically via `nft -f`.
- Tier order: (1) deny, (2) reply-stamp restore, (3) stamp specific, (4) stamp fallthrough (`from_entity_world`), (5) est/rel accept, (6) drop. Position is by action + specificity, never creation order.
- Within tiers 3/4, same direction+ports are sorted specific-before-fallthrough with `created_at` as tiebreaker; two overlapping specific policies (same dir+ports, both `--cidrs`) are rejected at create.
- Registry files are never written by the daemon poll loop.

brunodam:
## Description

Story 1.5 of epic TS_1 (#738) — "Policy registry and tier-order chain re-render." Most of the acceptance criteria already shipped as part of Story 1.2 (`network policy create`, PR #803): the JSON registry at `/etc/solo-provisioner/policies/<name>.json`, the atomic full-chain re-render of `inet weaver` in tier order (deny → reply-restore → stamp-specific → stamp-fallthrough → est/rel → drop) applied via `nft -f`, and the daemon poll loop never writing the registry. This PR closes the two acceptance criteria that weren't yet implemented:

- **Overlap rejection** — `network policy create` now rejects a "specific" `--stamp` policy (one that is not `--from-entity world`) that shares its `(Direction, Ports)` with another already-registered specific policy. Two such policies would have an ambiguous classification outcome, since set membership can diverge independently after create via the daemon/element verbs (Story 1.3), so the check is on the group key alone, not the initial `--cidrs` values.
- **`created_at` tiebreak** — within tiers 3/4, rules for policies that share a `(Direction, Ports)` group now render in creation order (`CreatedAt` ascending) instead of falling back to alphabetical-by-name order.

Fallthrough (`--from-entity world`) policies are exempt from the overlap check by design — they intentionally match anything, and the `created_at` tiebreak already gives them a deterministic position.

### Files changed

| File | Change |
|---|---|
| `internal/network/policy/policy.go` | Add `isSpecific`, `portsKey`, `groupKey`, `checkNoOverlap` helpers |
| `internal/network/policy/manager.go` | Wire `checkNoOverlap` into `Manager.Create`, before any render/kernel/disk mutation |
| `internal/network/policy/render.go` | `renderChain`'s tier-3/4 loop groups by `(Direction, Ports)` and stably sorts each group by `CreatedAt` via new `orderByGroupThenCreatedAt` |
| `internal/network/policy/policy_test.go` | New tests: overlap rejection, self-exclusion on `--force`, fallthrough exemption, `created_at` tiebreak |

### Review guide

**Code review checklist**

- `checkNoOverlap` only fires for `Action == ActionStamp && !FromEntityWorld` pairs — a `--deny` policy or a `--from-entity world` policy sharing the same direction+ports as another policy is never rejected.
- `checkNoOverlap` skips the registry entry whose `Name` matches the candidate, so a `--force` re-create of the same policy never self-rejects.
- `portsKey` sorts a copy of `--ports` before joining, so port order in the flag doesn't affect either overlap detection or tier grouping.
- `orderByGroupThenCreatedAt`'s per-group sort is `sort.SliceStable` — verified against the existing golden fixture (`TestRender_GoldenMatchesBNInstallSet`): none of its specific policies share a `(Direction, Ports)` tuple today, so the new grouping is a no-op there and the test passes unmodified.
- The overlap-rejection error uses `errorx.IllegalArgument` (operator-actionable — rename or adjust one of the two conflicting policies).
- `checkNoOverlap` runs before `snapshotMembership`/`Render`/`Apply` in `Manager.Create` — a rejected overlap never touches the kernel or the registry file.

**Test commands**

```bash
# Unit tests (macOS OK — no Linux-only deps in internal/network/policy)
go test -race -cover -tags='!integration' ./internal/network/policy/...

# Just the new/changed tests
go test ./internal/network/policy/... -run 'TestCreate_Overlap|TestCreate_FromEntityWorldNotSubjectToOverlapCheck|TestRender_CreatedAtTiebreak' -v

# Lint
task lint:check
```

**Manual UAT** (Linux box/VM with `nft` and an existing `inet weaver` table)

1. Create two specific policies with the same direction+ports:
   ```bash
   solo-provisioner network policy create --name test-a --stamp publisher --ports 40840 --cidrs 10.1.0.1/32
   solo-provisioner network policy create --name test-b --stamp reserve-ingress --ports 40840 --cidrs 10.1.0.2/32
   ```
   Expected: the second command fails with `policy "test-b" overlaps with existing policy "test-a": both are specific --stamp policies for direction=ingress ports=[40840]`; `nft list table inet weaver` shows no `test-b` set.

2. Create two fallthrough policies with the same direction+ports (should succeed):
   ```bash
   solo-provisioner network policy create --name test-c --stamp public --from-entity world --ports 9000
   solo-provisioner network policy create --name test-d --stamp reserve-egress --from-entity world --ports 9000
   ```
   Expected: both succeed; `cat /etc/solo-provisioner/network-weaver.nft` shows `test-c`'s rule before `test-d`'s in the fallthrough block.

**Risks / rollback**

- The overlap check is scoped narrowly (`Action == ActionStamp && !FromEntityWorld`); if a legitimate use case surfaces that it wrongly blocks, it can be relaxed in `checkNoOverlap`/`isSpecific` without touching `renderChain` or the `Manager.Create` call site.
- The tier-3/4 grouping change and the overlap check are independent and can be reverted separately.
- No registry file format change — existing registry JSON parses unchanged; the tiebreak only affects in-memory render ordering.

### Related Issues

* Closes #772

### DIFF
--- a/internal/network/policy/manager.go
+++ b/internal/network/policy/manager.go
@@ -127,6 +127,14 @@ func (m *Manager) Create(ctx context.Context, p *Policy, cidrs []string, podCIDR
 			}
 		}
 
+		// Reject a specific --stamp policy that would share its
+		// (Direction, Ports) group with another already-registered
+		// specific policy -- checked up front, for both the
+		// create-if-missing and --force paths.
+		if err := checkNoOverlap(policies, p); err != nil {
+			return err
+		}
+
 		existing := findByName(policies, p.Name)
 		target, newCIDRs := p, cidrs
 		if existing != nil && !force {
--- a/internal/network/policy/policy.go
+++ b/internal/network/policy/policy.go
@@ -4,6 +4,7 @@ package policy
 
 import (
 	"net"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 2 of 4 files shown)


alex-au (approved):
Nice and clean!

PR #909, opened 2026-07-21
feat(blocknode): orchestrate network policy plane + daemon setup at install

### LINKED ISSUE #762: Story 2.3 — `block node install` orchestrates the static plane via `network … create`
**Epic:** #735 — Static-plane install & reboot persistence

Design: [v4 design](https://github.com/automa-saga/traffic-shaper/blob/main/docs/bn-qos-multiclass-priority-design-v4-nft-priority.md) §8.2.

Implement the one-time install that lays down the BN **workload** static plane by delegating to the `network` subcommands (TS_1). The install is a thin orchestrator; the primitives live in TS_1.

The host firewall (`inet host`: SSH/mgmt/ICMP/in-cluster ports) is **not** created here — it is node-agnostic and owned by cluster/host provisioning (#777), applied for all node types at `kube cluster install`. `block node install` assumes it is already present.

**Acceptance criteria:**
- Required `--egress-bandwidth <speed>`; optional `--ingress-bandwidth` (defaults to egress) and per-class `--shape <class>=rate=...,ceil=...,prio=...` overrides forwarded to `network shape create`.
- Runs idempotently (create-if-missing): `network shape create` per class then device roots (classes first so `--default` resolves); `network policy create` per category (bn-publisher, bn-subscriber-in, bn-partner-out, bn-public-out, bn-status-in/out, bn-mgmt-in/out on the BN admin port 40983, bn-restricted, bn-backfill); `network lb create`.
- Does **not** invoke `network firewall create` (that is the host layer's job, #778).
- Re-running does not clobber operator mutations to set membership or per-class shape; `--force` re-renders static rules + tc script.
- Operator-curated sets (bn-mgmt, bn-restricted) receive initial membership via `--cidrs` at `policy create`.

### LINKED ISSUE #798: `block node install` should remind operator to run `daemon service install` when daemon is not yet active
## Problem

`solo-provisioner block node install` completes successfully but prints no guidance about
the daemon. A node operator who wants traffic shaping (or any future daemon-backed feature)
has no in-tool signal that they need to also run `sudo solo-provisioner daemon service install`.

The two binaries are released and installed independently, and the daemon's `daemon.yaml`
config (including `monitors.traffic_shaper: true`) is only written during `daemon service install`.
Without it the traffic-shaper monitor never runs, even after the block node is live — and there
is currently nothing in the `block node install` output that tells the operator this.

## Proposed fix

At the end of a successful `block node install` workflow, check whether
`solo-provisioner-daemon.service` is active (e.g. `systemctl is-active`) or whether
`daemon.yaml` is present at the expected config path. If not, emit a single notice-level
message in the post-workflow summary, for example:

```
Notice: the solo-provisioner daemon is not installed on this host.
To enable traffic shaping and other daemon-backed features, run:
  sudo solo-provisioner daemon service install --block-node-enabled
```

No failure, no blocking — purely informational. The check should be a no-op on hosts
where the daemon is already running.

## Acceptance

- [ ] After a successful `block node install`, if the daemon service is not active, a notice is printed pointing to `daemon service install`.
- [ ] If the daemon is already running (service active), no notice is printed.
- [ ] The check adds no meaningful latency to the install workflow.

### LINKED ISSUE #904: `block node install` should offer to install + provision the traffic-shaper daemon
## Problem

Installing a block node and getting its traffic-shaper daemon running are two separate operator steps today: `solo-provisioner block node install`, then a separate `solo-provisioner daemon service install --components block-node --bn-orbit <ns>`. The daemon step is easy to forget, and a block node without its traffic-shaper daemon silently has no ingress prioritization (the veth HTB is never installed). #798 covers *reminding* the operator after the fact; this issue proposes closing the gap in `block node install` itself.

## Proposed fix

Have `block node install` offer to install + provision the traffic-shaper daemon as part of the flow:

- Interactive: a TUI prompt ("Install and enable the block-node traffic-shaper daemon now?") that, on yes, runs the daemon install + provisioning (RBAC + `daemon-bn.kubeconfig` + systemd unit) scoped to the block node's namespace.
- Non-interactive: a flag (e.g. `--install-daemon` / `--with-daemon`) to do the same without prompting; omitted → behave as today and print the reminder from #798.

Depends on block-node daemon provisioning existing (the RBAC + kubeconfig step) — see the companion provisioning story.

## Acceptance

- [ ] Interactive `block node install` offers to install + provision the traffic-shaper daemon; accepting leaves the daemon running with the traffic-shaper monitor active.
- [ ] A non-interactive flag drives the same behavior without prompting.
- [ ] Declining (or omitting the flag) falls back to the #798 reminder; no daemon is installed.
- [ ] The namespace/orbit is derived from the block node install inputs, not re-prompted.

### LINKED ISSUE #915: block node install unconditionally installs the tc-egress systemd service — breaks systemd-less environments (CI/kind)
## Summary

As of **v0.22.0**, `block node install` unconditionally runs the traffic-shaper `TcEgressPersist`
step, which renders and **installs a systemd service unit** (`tc-egress.service`) for the `$EGRESS`
HTB hierarchy. In an environment without systemd/dbus (e.g. a containerized CI runner, or kind on
such a runner) this aborts the whole install:

```
Failed to persist tc-egress hierarchy
  cause: failed to install tc-egress service unit
  cause: os.systemd_connection_error: failed to connect to systemd,
         cause: dial unix /var/run/dbus/system_bus_socket: connect: no such file or directory
✗ Error: failed to execute command
```

## Where (v0.24.0)

- `internal/bll/blocknode/install_handler.go` — both install branches call the network-setup phase
  that includes `steps.TcEgressPersist(ins.EgressInterface, ins.LinkRate)` unconditionally (line ~85).
- `internal/network/shape/service_linux.go` — installs/enables `tc-egress.service` via systemd.

There is no flag or config to skip it. Introduced in v0.22.0; v0.20.0/v0.21.0 install has no
tc-egress step and works in these environments.

## Impact

A downstream E2E lifecycle test (hiero-block-node's `e2e-node-operator` workflow) runs
`solo-provisioner block node install` on a kind cluster on an ARC (containerized) GitHub runner with
no systemd. On v0.21.0 it passes; bumping to v0.22.0+ fails at install on the tc-egress step. This
blocks adopting v0.22.0+ — including the reset-storage behavior we want — so the test is pinned back
to v0.21.0.

## Request

Add a way to skip the traffic-shaper / tc-egress install (e.g. a `--no-traffic-shaper` /
`--skip-network-setup` flag, a config field, or auto-skip when systemd is unavailable), so
`block node install` succeeds in systemd-less environments. The shaper is a host-network optimization,
not required for a functional Block Node in a test/CI cluster.

## Suggested triage

`Bug` or `Feature Enhancement`, `BN`. (Labels not set here — filer has read-only access.)

brunodam:
## Description

`block node install` is now a complete, opt-in static-plane orchestrator that also
closes the daemon-activation gap. This PR grew across several rounds as review and
manual UAT surfaced real gaps in the original design; this description reflects the
final state, not the chronological history.

### What `block node install` does now

- **Host firewall** (`--firewall-enabled`, default `false`, opt-in): applies the
  node-level `inet host` firewall (SSH/mgmt allowlist, ICMP policy, in-cluster
  ports). `--mgmt-cidrs` configures the allowlist; the new `--blocked-cidrs` adds a
  permanent, purely operator-managed deny list (set `@blocked_addrs`, dropped before
  every other rule including established connections) — distinct from the BN
  workload plane's `bn-restricted` set, which is daemon-owned (see below).
- **Traffic shaping** (`--traffic-shaping-enabled`, default `false`, opt-in): one
  gate that creates the BN workload network-policy plane (`inet weaver`
  classification, the fixed set of BN policies), tc HTB shaping for `$EGRESS`/`$VETH`,
  and installs the traffic-shaper daemon — all three together, no separate daemon
  prompt or flag. Declining skips the egress NIC/link-rate prompts too. This decision
  is durable: it's recorded in `state.yaml` (`BlockNodeState.TrafficShapingDisabled`)
  and honored by later `reconfigure`/`upgrade` runs, not just the install where it
  was made.
- **Daemon binary source** (`--daemon-version`, `--daemon-bin`): the daemon binary is
  auto-downloaded from the infrastructure catalog at `--daemon-version` (defaults to
  this CLI's own version) unless `--daemon-bin` points at a local binary instead.
  `--daemon-bin` is required for `--profile=local` (prompted interactively, or a
  fail-fast error non-interactively) since local/dev builds have no downloadable
  release.
- Adds per-class `--shape <class>=rate=...,ceil=...,prio=...` overrides.

### Why the defaults are opt-in (`false`), not opt-out

An earlier version of this PR defaulted both gates to `true`. Two things forced the
change: (1) it would have been a breaking change for every existing non-interactive/
scripted caller of `block node install` that doesn't pass these flags — they'd
silently start getting a host firewall, network policy plane, tc shaping, and a
daemon install they never asked for; (2) the daemon auto-install path is currently
broken regardless (see below), so defaulting it on would make a plain `block node
install` fail by default. Opt-in avoids both problems until these features have more
mileage.

### Known follow-up (not in this PR): daemon catalog entry

`pkg/software/infrastructure-catalog.yaml` has no entry for
`solo-provisioner-daemon`, so `--daemon-version` auto-download currently fails with
`software 'solo-provisioner-daemon' not found in configuration`. This is a
pre-existing gap (not introduced here) — tracked separately; `--daemon-bin` is the
workaround until it's added.

### Known limitation (not in this PR): reconfigure/upgrade can't change either gate

Both gates are effectively install-time-only decisions. `reconfigure` happens to
be able to *enable* the host firewall later (its prompt/flags re-resolve fresh on
every run and `NetworkFirewallCreate` is create-if-missing), but there's no
equivalent for traffic shaping, and neither command can *disable* anything —
there's no teardown logic outside a full `block node uninstall`. For an
already-installed block node, the only way to turn either feature on or off today
is a full `block node install --force` reinstall. Filed as #911.

### Bug fixed along the way: effective-inputs passthrough

`resolveBlocknodeEffectiveInputs` (`internal/bll/blocknode/helpers.go`) — whose
output is what actually reaches `BuildWorkflow` — was missing `ShapeOverrides` and
`TrafficShapingEnabled` from its "passed through from user input" section, silently
dropping `--shape` and the traffic-shaping gate's resolved value on every install.
Fixed.

### CI fix: unclassified pod egress (Maven plugin resolution)

UAT `block node upgrade --chart-version 0.29.0` was crash-looping the chart's
`resolve-plugins` init container. Root cause: the `inet weaver` forward chain only
allowed BN's own ports, with no rule for the chart's Maven-based plugin-prefix
resolution (repo1.maven.org, hyperledger.jfrog.io, artifacts.consensys.net,
central.sonatype.com on 443) — never surfaced at the initial 0.26.2 install since
that chart feature is gated to >=0.28.1. Fixed with a new unclassified-pod-egress
tier: a blanket `ip saddr <podCIDR> ip daddr != <podCIDR> accept` with no `meta
priority`, so unclassified outbound traffic falls to the HTB default class (shaped,
not prioritized) instead of being dropped. This is a deliberate default-allow escape
hatch — see Risks.

## Files changed

| File | Change |
|---|---|
| `internal/network/shape/override.go` (new) | `ClassOverride`, validation, apply, sum-check |
| `internal/network/shape/manager.go` | `ProvisionDefault{Egress,Ingress}` accept overrides |
| `internal/workflows/steps/step_network_policy.go` (new) | `NetworkPolicyCreate` step, canonical BN policy set (`bn-mgmt-in/out` curated from `--mgmt-cidrs`; `bn-restricted` always empty, daemon-owned) |
| `internal/workflows/steps/step_network_tc_{egress,ingress}.go` | Accept + forward `--shape` overrides |
| `internal/network/policy/render.go` | New tier: unclassified pod egress (Maven-egress fix) |
| `internal/network/firewall/{table,render,parse,manager}.go` | `BlockedCIDRs` field, `@blocked_addrs` set, `Add/Remove/SetBlockedCIDR(s)` |
| `internal/templates/files/network/network-host.nft.tmpl` | `@blocked_addrs` set + early drop rule (before established/related) |
| `internal/workflows/steps/step_network_firewall.go` | Applies `hostCfg.BlockedCIDRs` |
| `internal/workflows/network_setup.go` | `trafficShapingEnabled` param gates `NetworkPolicyCreate`/`NftWeaverPersist`/`TcEgressPersist`/`TcIngressRecord` as one bundle |
| `internal/state/state.go` | `BlockNodeState.TrafficShapingDisabled` (persisted, negative polarity) |
| `internal/bll/blocknode/install_handler.go` | Threads `trafficShapingEnabled`; `patchBlockNodeStateAfterInstall()`; daemon-config workflow conditional |
| `internal/bll/blocknode/reconfigure_handler.go` | Reads `currentState.BlockNodeState.TrafficShapingDisabled`; omits `TcEgressPersist` across all 4 branches when disabled |
| `internal/bll/blocknode/helpers.go` | Fixed `ShapeOverrides`/`TrafficShapingEnabled` passthrough; new `patchBlockNodeStateAfterInstall()` (install-only) |
| `internal/workflows/steps/step_daemon.go` | `DaemonBinarySource.Version` field, wired to `software.WithVersion()` |
| `pkg/models/inputs.go` | `ShapeOverride` type; `ShapeOverrides`/`TrafficShapingEnabled` fields |
| `pkg/models/config.go` | `HostConfig.BlockedCIDRs` field + validation |
| `cmd/cli/commands/common/host_firewall.go` | `--blocked-cidrs` flag; `--firewall-enabled` default flipped to `false` |
| `cmd/cli/commands/common/traffic_shaping.go` (new) | `--traffic-shaping-enabled` flag (default `false`) + gate resolution/prompt |
| `cmd/cli/commands/common/flags_common.go` | `FlagDaemonVersion()` (new) |
| `cmd/cli/commands/block/node/install.go` | `--shape`, `--traffic-shaping-enabled`, `--daemon-version`, `--daemon-bin`; gates egress prompts/daemon-source resolution/daemon activation on the traffic-shaping gate |
| `cmd/cli/commands/block/node/daemon_offer.go` | `ensureBlockNodeDaemon` (renamed from `offerDaemonInstall`, no prompt — installs unconditionally when called); `resolveDaemonBinarySource` (local-profile `--daemon-bin` requirement) |
| `cmd/cli/commands/network/firewall/*.go` | `--blocked-cidrs`/`--blocked-cidr` across `create`/`add`/`remove`/`set` |
| `internal/ui/prompt/cluster.go` | `BlockedCIDRsInputPrompt` |
| `docs/quickstart.md` | All new flags + notes; removed flags dropped |

## Test plan

```bash
# Native (macOS-safe) — pure logic
go test ./internal/network/shape/... ./internal/network/policy/... ./internal/network/firewall/... ./internal/ui/prompt/... ./pkg/models/...

# Full unit suite (Linux-only packages: steps, common, bll) — run in UTM VM
task vm:test:unit

# Cross-compile check (validates Linux-target type-correctness without a VM)
GOOS=linux GOARCH=amd64 go vet ./...

# Lint (0 issues expected)
task lint
```

## Manual UAT

Steps 1–11 are single-host checks with the actual commands. Steps 12–13 use the
accessory-VM IPs: step 12 drives the daemon-owned nft sets from a mock statusz
roster keyed on those IPs (via the `reconcile-shaper` worker), and step 13 is the
deeper multi-VM traffic validation — two accessory VMs generating concurrent
traffic while a `tc_rates` helper streams live per-class Mbit/s on the BN-host,
proving priority preemption holds under contention (not just that the tc classes
exist).

1. **Default install, non-interactive** — nothing network-related gets created:
   ```bash
   sudo solo-provisioner block node install --profile=local --non-interactive
   sudo nft list tables | grep -E 'inet (host|weaver)'   # expect: no output
   systemctl is-active solo-provisioner-daemon            # expect: inactive
   ```
2. **Traffic shaping opt-in**:
   ```bash
   sudo solo-provisioner block node install --profile=local \
     --traffic-shaping-enabled --daemon-bin <path> --non-interactive
   sudo nft list table inet weaver | head -20   # 10 canonical policies
   sudo solo-provisioner network shape show     # 6 classes, rate/ceil/prio
   systemctl is-active solo-provisioner-daemon  # active, no separate prompt/flag needed
   ```
3. **Host firewall opt-in**:
   ```bash
   sudo solo-provisioner block node install --profile=local \
     --firewall-enabled --mgmt-cidrs 10.0.0.0/8 --non-interactive
   sudo nft list table inet host   # default-drop, @mgmt_addrs = {10.0.0.0/8}
   ```
4. **Host firewall block list** — quarantine proven with a live connection:
   ```bash
   sudo solo-provisioner network firewall create --blocked-cidrs 198.51.100.0/24
   # from 198.51.100.x: nc -zv <bn-host-ip> 22 → refused/timeout within ~1s
   sudo nft list table inet host | grep blocked_addrs   # elements present
   sudo solo-provisioner network firewall add/remove --blocked-cidr ...   # round-trips
   ```
5. **Shape override** — `publisher` is an ingress class (classid `1:10`), so it
   lives on `$VETH` (the BN pod's host-side veth peer), not the physical
   `$EGRESS` NIC. Resolve `$VETH` the same way the daemon does: match the pod's
   `eth0` iflink against the host's `/sys/class/net/*/ifindex`. The block node
   installs into the `block-node` namespace by default (`--namespace` to
   override):
   ```bash
   BN_POD=$(kubectl get pod -n block-node -l app.kubernetes.io/name=block-node-server -o jsonpath='{.items[0].metadata.name}')
   IFLINK=$(kubectl exec -n block-node "$BN_POD" -- cat /sys/class/net/eth0/iflink)
   VETH=$(for f in /sys/class/net/*/ifindex; do [ "$(cat "$f")" = "$IFLINK" ] && basename "$(dirname "$f")"; done)

   sudo solo-provisioner block node install --profile=local --traffic-shaping-enabled \
     --daemon-bin <path> --shape publisher=rate=800mbit,ceil=1gbit,prio=0 --non-interactive
   sudo solo-provisioner network shape show   # publisher: rate=800mbit ceil=1gbit prio=0
   CLASSID=$(sudo solo-provisioner network shape show | awk -F'[()]' '/publisher/{print $2}')  # e.g. 1:10
   sudo tc class show dev "$VETH" classid "$CLASSID"  # htb rate 800mbit ceil 1gbit
   ```
6. **Durability across reconfigure** — declined stays declined. `$EGRESS` is the
   physical NIC auto-detected from the default route (same detection
   `TcEgressPersist` uses internally):
   ```bash
   EGRESS=$(ip route get 8.8.8.8 | grep -oP 'dev \K\S+')

   sudo solo-provisioner block node install --profile=local --non-interactive   # declined (default)
   sudo solo-provisioner block node reconfigure --recent-retention 50000
   sudo tc qdisc show dev "$EGRESS"   # expect: still no htb qdisc (was silently re-enabled pre-fix)
   ```
7. **Upgrade past 0.28.1**:
   ```bash
   sudo solo-provisioner block node upgrade --profile=local --chart-version 0.29.0
   kubectl logs -n block-node -l app.kubernetes.io/name=block-node-server -c resolve-plugins --tail=20
   # expect: succeeds, no crash loop
   BN_POD=$(kubectl get pod -n block-node -l app.kubernetes.io/name=block-node-server -o jsonpath='{.items[0].metadata.name}')
   kubectl exec -n block-node "$BN_POD" -- curl -sSI --max-time 5 https://repo1.maven.org/maven2/ | head -1
   # expect: HTTP response — proves the unclassified-pod-egress tier
   ```
8. **`--profile=local` without `--daemon-bin`**:
   ```bash
   sudo solo-provisioner block node install --profile=local --traffic-shaping-enabled --non-interactive
   # expect: "--daemon-bin is required for --profile=local ..."
   ```
9. **Host-firewall flags without the gate, non-interactive**:
   ```bash
   sudo solo-provisioner block node install --profile=local --non-interactive --mgmt-cidrs 10.0.0.0/8
   # expect: "--mgmt-cidrs was supplied but the host firewall is not enabled ..."
   ```
10. **Traffic-shaping flags without the gate, non-interactive**:
    ```bash
    sudo solo-provisioner block node install --profile=local --non-interactive \
      --shape publisher=rate=800mbit,ceil=1gbit,prio=0
    # expect: "--shape was supplied but traffic shaping is not enabled ..."
    ```
11. **`--shape` error message names the class**:
    ```bash
    sudo solo-provisioner block node install --profile=local --traffic-shaping-enabled \
      --daemon-bin <path> --shape publisher=rate=bogus --non-interactive
    # expect: error mentions "publisher" and "--shape rate" (not a bare parse error)
    ```
12. **Daemon-owned nft sets reconcile from statusz (mock roster)** — proves the
    daemon-owned `inet weaver` sets (`bn-publisher`, `bn-partner-out`,
    `bn-restricted`, and the compound `bn-backfill`) follow the block node's
    reported endpoints. The daemon's in-process poll loop is still an idle stub
    (`statusz poll loop not yet wired — reconcile-shaper scheduler pending`), so a
    daemon restart alone reconciles nothing; membership is driven by running the
    `reconcile-shaper` worker by hand against a mock statusz server (the same
    worker the scheduler will eventually exec). Requires a prior
    `--traffic-shaping-enabled` install (step 2) so the sets already exist. The
    mock re-reads its roster file on every request, so edits take effect on the
    next run with no restart. Seed the roster from the accessory-VM IPs
    (`$CN_VM_IP` publisher, `$PARTNER_VM_IP` partner, `$PEER_VM_IP` backfill peer);
    `$PUBLIC_VM_IP` deliberately has no daemon-owned set — `bn-public-out` is a
    `--from-entity world` port match, not statusz-reconciled.
    ```bash
    cd /mnt/solo-weaver
    cat > /tmp/roster.json <<EOF
    {
      "inbound": [
        { "remote": {"address": "${CN_VM_IP}/32", "port": "*"}, "category": "publisher" },
        { "remote": {"address": "${PARTNER_VM_IP}/32", "port": "*"}, "category": "partner" }
      ],
      "outbound": [
        { "remote": {"address": "${PEER_VM_IP}", "port": "43473"}, "category": "peer_bn" }
      ]
    }
    EOF
    nohup go run ./internal/daemon/blocknode/statuszmock/cmd \
      --addr 127.0.0.1:8080 --roster /tmp/roster.json > /tmp/statuszmock.log 2>&1 &
    ```
    `peer_bn` is a compound (address·port) set, so the backfill peer needs a
    concrete non-wildcard port (the port the BN's backfill client dials on the
    peer; wildcard/empty ports are skipped). Then reconcile:
    ```bash
    # dry-run first (unprivileged) — fetches statusz, prints the desired-membership
    # digest, touches no nft state:
    solo-provisioner block node reconcile-shaper --statusz-url http://127.0.0.1:8080 --check
    # apply (root) — diffs live nft and rewrites only the changed policy sets:
    sudo solo-provisioner block node reconcile-shaper --statusz-url http://127.0.0.1:8080
    # expect: applied: bn-backfill, bn-partner-out, bn-publisher

    sudo nft list set inet weaver bn-publisher    # elements = { $CN_VM_IP }
    sudo nft list set inet weaver bn-partner-out  # elements = { $PARTNER_VM_IP } — the bn-partner-out fix
    sudo nft list set inet weaver bn-backfill     # elements = { $PEER_VM_IP . 43473 } — compound addr . port
    sudo nft list set inet weaver bn-public-out   # no elements clause (port match, not statusz-driven)
    ```
    Drop the partner entry from `/tmp/roster.json`, re-run the apply, and confirm
    `bn-partner-out` clears on the next reconcile — validates the "category
    present, membership empty → set cleared" contract.
13. **Priority preemption across VMs (`tc_rates`)** — proves shaping actually
    enforces priority under contention. **Two parts,** because the real Block Node
    chart is not an iperf3 server and its LoadBalancer Service exposes only
    `40840`; the facility ports `40980`–`40983` live only in the host nft policy
    plane this PR adds, not on the Service. So `iperf3 -c $BN_LB_IP` hangs against
    the real chart — 40840's TCP handshake succeeds but the listener is the BN
    gRPC server, not `iperf3 -s`, and 40980/40981 aren't served at all.

    **(a) Classification-only, against the real chart (runnable now)** — proves the
    SYN from a roster-classified source gets `meta priority` marked; a bare TCP
    connect suffices (the app may RST the unknown protocol right after —
    classification happens on the SYN, before any L7 payload):
    ```bash
    # 🐧 BN-host: trace the publisher rule, then connect from the CN VM
    sudo nft add rule inet weaver forward ip saddr $CN_VM_IP tcp dport 40840 meta nftrace set 1
    sudo nft monitor trace & TRACE_PID=$!
    # 🟦 CN VM:  nc -zv $BN_LB_IP 40840
    kill $TRACE_PID   # trace shows the rule firing → meta priority 0x10010 on the SYN
    ```

    **(b) Sustained throughput/preemption (optional — needs the iperf3 stand-in)** —
    validates the shaping *mechanism* (already proven upstream), so skip it if (a)
    is enough. `iperf3` needs a real `iperf3 -s` behind each port, which the real
    chart does not run: swap it for the POC's Alpine+iperf3 stand-in
    Deployment+Service (exposes all facility ports, runs `iperf3 -s` on each) —
    `block node uninstall` leaves the `inet weaver`/tc plane in place — then run the
    live per-class stream below.
    Requires two accessory VMs (e.g. PARTNER + PUBLIC, cloned from the same
    golden image as the BN-host) reachable from it. On the BN-host, resolve the
    MetalLB-assigned LB IP once and pass it to both accessory VMs as
    `$BN_LB_IP` (exact Service name depends on the Helm release name used):
    ```bash
    kubectl get svc -n block-node -o json | jq -r '.items[] | select(.spec.type=="LoadBalancer") | select(.metadata.name | test("block-node")) | .status.loadBalancer.ingress[0].ip'
    ```
    ```bash
    # 🐧 BN-host VM — stream live per-class Mbit/s (Ctrl-C to stop)
    tc_rates() {
        local dev=$1 classes=$2 interval=${3:-2}
        (while true; do sudo tc -s class show dev "$dev"; echo "TICK $(date +%s.%N)"; sleep "$interval"; done) \
        | mawk -v CLASSES="$classes" '
    BEGIN { n=split(CLASSES,cls,","); for(i=1;i<=n;i++) want[cls[i]]=1 }
    /^class htb / { c=(want[$3])?$3:"" }
    /Sent/&&c { cur[c]=$2; c="" }
    /^TICK / { now=$2+0; if(warmed){dt=now-last_t; for(i=1;i<=n;i++){k=cls[i]; d=cur[k]-prev[k];
      printf "  %-5s rate=%6.2f Mbit/s\n", k, (d*8)/(dt*1000000)}} else warmed=1
      for(i=1;i<=n;i++) prev[cls[i]]=cur[cls[i]]; last_t=now }'
    }
    EGRESS=$(ip route get 8.8.8.8 | grep -oP 'dev \K\S+')   # see step 6 if already resolved
    PARTNER_CLASS=$(sudo solo-provisioner network shape show | awk -F'[()]' '/partner/{print $2}')
    PUBLIC_CLASS=$(sudo solo-provisioner network shape show | awk -F'[()]' '/public/{print $2}')
    tc_rates "$EGRESS" "$PARTNER_CLASS,$PUBLIC_CLASS" 2
    ```
    ```bash
    # 🟩 PARTNER VM (higher priority, e.g. prio=0) — start first
    iperf3 -c $BN_LB_IP -p 40980 -t 60 -R
    ```
    ```bash
    # 🟨 PUBLIC VM (lower priority, e.g. prio=5) — start ~5s later, overlapping
    iperf3 -c $BN_LB_IP -p 40981 -t 45 -R
    ```
    Expected: partner holds its configured `rate` floor throughout the overlap;
    public yields bandwidth while partner is active and recovers toward its
    `ceil` within ~5s once partner's transfer ends.

## Risks / rollback

- **Default-allow escape hatch.** The unclassified-pod-egress tier broadens BN pod
  outbound reachability beyond BN's own ports (unclassified traffic is shaped, not
  blocked). Deliberate trade-off to unblock the chart's plugin resolution; flagged
  here per the design discussion that led to it.
- **Daemon auto-download is currently non-functional** (missing catalog entry, not
  introduced by this PR) — `--daemon-bin` is the workaround until a follow-up adds
  the catalog entry.
- **The daemon's statusz poll loop is a stub in this PR.** The reconcile engine,
  the `reconcile-shaper` worker, and the `statusz.base_url`/`poll_interval` config
  all land here, but `TrafficShaperMonitor.runStatuszPoll` is deliberately an idle
  no-op — the scheduler that execs the worker on each poll tick is not yet wired.
  A running daemon therefore does not reconcile the daemon-owned `inet weaver` sets
  (`bn-publisher`, `bn-partner-out`, `bn-restricted`, `bn-backfill`) on its own;
  reconciliation is driven by running the worker by hand until the scheduler lands.
  Filed as #914.
- **Defaults changed from an earlier revision of this same PR** (not from `main`):
  `--firewall-enabled` and `--traffic-shaping-enabled` now default to `false`. Since
  neither flag exists on `main` yet, this is not a breaking change relative to
  released behavior.
- **`--restricted-cidrs` was added then removed within this PR** (never released) —
  `bn-restricted` is confirmed daemon-owned (reconciled from BN's statusz every poll
  tick), so an install-time seed would have been silently overwritten; no user-facing
  regression since the flag never shipped.
- **No migration boundary.** All of this is install-time orchestration; it does not
  touch embedded templates or startup migrations, so none of it can leak into
  already-provisioned clusters via the upgrade path.
- **Rollback.** Revert the commit(s); no on-disk schema changed other than the new
  `state.yaml` field (`trafficShapingDisabled`), which is additive and ignored by
  older binaries.

## Related Issues

* Closes #762
* Closes #904
* Closes #798
* Closes #915 
* Follow-up filed: #911 (reconfigure/upgrade can't enable or disable either gate on an existing install)
* Follow-up filed: #914 (wire the daemon-side statusz poll scheduler that execs `reconcile-shaper` on an interval)

### DIFF
--- a/cmd/cli/commands/block/node/daemon_offer.go
+++ b/cmd/cli/commands/block/node/daemon_offer.go
@@ -1,0 +1,170 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package node
+
+import (
+	"github.com/automa-saga/logx"
+	"github.com/hashgraph/solo-weaver/cmd/cli/commands/common"
+	"github.com/hashgraph/solo-weaver/internal/daemon"
+	"github.com/hashgraph/solo-weaver/internal/ui/prompt"
+	"github.com/hashgraph/solo-weaver/internal/workflows"
+	workflowsteps "github.com/hashgraph/solo-weaver/internal/workflows/steps"
+	"github.com/hashgraph/solo-weaver/pkg/models"
+	pkgos "github.com/hashgraph/solo-weaver/pkg/os"
+	"github.com/hashgraph/solo-weaver/pkg/sanity"
+	"github.com/joomcode/errorx"
+	"github.com/spf13/cobra"
+)
+
+// daemonServiceName is the systemd unit for the solo-provisioner daemon; matched
+// against systemd's active state to decide whether the daemon is already running.
+const daemonServiceName = "solo-provisioner-daemon"
+
+// ensureBlockNodeDaemon closes the gap between installing a block node and
+// getting its traffic-shaper daemon running. Daemon activation is not a
+// separate decision from the operator: it is part of the traffic-shaping
+// bundle. install.go calls this only when TrafficShapingEnabled is true — a
+// block node with the policy plane and tc shaping in place, but no daemon,
+// would otherwise have no ingress prioritization (the veth HTB is never
+// installed) and nothing reconciling the daemon-owned nft sets from statusz.
+//
+// When the daemon is already active there is nothing to do; otherwise it is
+// installed and provisioned (RBAC + daemon-bn.kubeconfig + systemd unit)
+// scoped to this block node's namespace, unconditionally — no prompt, no flag.
+// source controls where the daemon binary itself comes from (see
+// resolveDaemonBinarySource).
+func ensureBlockNodeDaemon(cmd *cobra.Command, namespace string, source workflowsteps.DaemonBinarySource) error {
+	ctx := cmd.Context()
+
+	// A running daemon needs no further action (the check is a no-op on hosts
+	// where the daemon is already up). A systemd/DBus error is non-fatal — treat
+	// the daemon as not running — but log it so an operator can tell a genuine
+	// "not installed" from a failed state query.
+	running, err := pkgos.IsServiceRunning(ctx, daemonServiceName)
+	if err != nil {
+		logx.As().Warn().Err(err).Msg(
+			"could not determine solo-provisioner-daemon service state; treating it as not running")
+	}
+	if running {
+		return nil
+	}
+
+	return provisionBlockNodeDaemon(cmd, namespace, source)
+}
+
+// resolveDaemonBinarySource determines where ensureBlockNodeDaemon should get
+// the daemon binary from: --daemon-bin (bypasses the catalog entirely) or
+// --daemon-version (selects which catalog version to auto-download; defaults
+// to this CLI's own version — meaningful once CLI and daemon are co-released
+// under one version scheme).
+//
+// --profile=local is a special case: local/dev builds report an unstamped
+// version (e.g. "dev") that has no corresponding catalog entry, so
+// auto-download can never succeed there. When traffic shaping is enabled and
+// --daemon-bin wasn't supplied, this becomes a required value — prompted for
+// interactively, or a fail-fast error non-interactively.
+func resolveDaemonBinarySource(cmd *cobra.Command, args []string, profile string, cv *prompt.ChosenValues) (workflowsteps.DaemonBinarySource, error) {
+	source := workflowsteps.DaemonBinarySource{BinPath: flagDaemonBin, Version: flagDaemonVersion}
+
+	if profile != models.ProfileLocal || source.BinPath != "" {
+		return source, nil
+	}
+
+	var rootFlags common.RootFlags
+	_ = common.ExtractRootFlags(cmd, args, &rootFlags)
+	if !prompt.ShouldPrompt(rootFlags.Force) {
+		return source, errorx.IllegalArgument.New(
+			"--daemon-bin is required for --profile=local: local/dev builds have no downloadable "+
+				"solo-provisioner-daemon release to auto-download").
+			WithProperty(models.ErrPropertyResolution, []string{
+				"Build the daemon locally: task build:daemon GOOS=linux GOARCH=<arch>",
+				"Then pass its path: --daemon-bin=<path-to-binary>",
+			})
+	}
+
+	localCV := cv
+	if localCV == nil {
+		localCV = prompt.NewChosenValues()
+	}
+	if err := prompt.RunInputPrompts(cmd, []prompt.InputPrompt{
+		daemonBinInputPrompt(source.BinPath, &source.BinPath),
+	}, localCV); err != nil {
+		return source, err
+	}
+	if source.BinPath == "" {
+		return source, errorx.IllegalArgument.New("--daemon-bin is required for --profile=local").
+			WithProperty(models.ErrPropertyResolution, []string{
+				"Build the daemon locally: task build:daemon GOOS=linux GOARCH=<arch>",
+				"Then pass its path: --daemon-bin=<path-to-binary>",
+			})
+	}
+	return source, nil
+}
+
+// daemonBinInputPrompt returns the interactive prompt for --daemon-bin, shown
+// only when --profile=local requires a local daemon binary path (see
+// resolveDaemonBinarySource).
+func daemonBinInputPrompt(eff string, target *string) prompt.InputPrompt {
+	return prompt.InputPrompt{
+		FlagName:       "daemon-bin",
+		Title:          "Daemon Binary Path (required for --profile=local)",
+		Description:    "Path to a locally-built solo-provisioner-daemon binary. Local/dev builds have no downloadable release to auto-download.",
+		Placeholder:    "bin/solo-provisioner-daemon-linux-amd64",
+		EffectiveValue: eff,
+		Target:         target,
+		Validate: func(s string) error {
+			if s == "" {
+				return errorx.IllegalArgument.New("daemon binary path cannot be empty for --profile=local")
+			}
+			_, err := sanity.SanitizePath(s)
+			return err
+		},
+	}
+}
+
+// provisionBlockNodeDaemon runs the daemon install + provisioning workflow for
+// the block-node component. daemon.yaml has already been written by the install
+// workflow's Traffic-shaper Monitor phase, so it is loaded here as the source of
+// truth; the block-node component's orbit is defaulted to this install's
+// namespace if not already set (never re-prompted).
+func provisionBlockNodeDaemon(cmd *cobra.Command, namespace string, source workflowsteps.DaemonBinarySource) error {
+	paths := models.Paths()
+
+	cfg, err := daemon.LoadDaemonConfig(paths.DaemonConfigPath)
+	if err != nil {
+		// A missing daemon.yaml is expected on some paths — fall back to a fresh
+		// config and provision for this namespace. A malformed (or otherwise
+		// unreadable) daemon.yaml is fatal: swallowing it would provision the
+		// daemon while the operator's broken config silently goes unnoticed.
+		if !daemon.IsConfigNotFound(err) {
+			return err
+		}
+		cfg = daemon.DaemonConfig{}
+	}
+	// This path is an explicit request to install/provision the block-node
+	// daemon, so force the block-node component and its traffic-shaper monitor
+	// on regardless of any pre-existing disabled state — otherwise
+	// NewDaemonServiceInstallWorkflow would skip BN RBAC/kubeconfig and the
+	// monitor would never run.
+	if cfg.Components.BlockNode == nil {
+		cfg.Components.BlockNode = &daemon.BlockNodeComponentConfig{}
+	}
+	cfg.Components.BlockNode.Enabled = true
+	cfg.Components.BlockNode.Monitors.TrafficShaper = true
+	if cfg.Components.BlockNode.Kubeconfig == "" {
+		cfg.Components.BlockNode.Kubeconfig = paths.DaemonBNKubeconfigPath
+	}
+	if cfg.Components.BlockNode.Orbit == "" {
+		cfg.Components.BlockNode.Orbit = namespace
+	}
+
+	wf, err := workflows.NewDaemonServiceInstallWorkflow(cfg, source)
+	if err != nil {
+		return err
+	}
+	if err := common.RunWorkflowBuilder(cmd.Context(), wf); err != nil {
+		return err
+	}
+	logx.As().Info().Msg("solo-provisioner-daemon service installed, enabled, and started")
+	return nil
+}
--- a/cmd/cli/commands/block/node/install.go
+++ b/cmd/cli/commands/block/node/install.go
@@ -8,6 +8,7 @@ import (
 	"github.com/automa-saga/automa"
 	"github.com/automa-saga/logx"
 	"github.com/hashgraph/solo-weaver/cmd/cli/commands/common"
+	workflowsteps "github.com/hashgraph/solo-weaver/internal/workflows/steps"
 	"github.com/hashgraph/solo-weaver/pkg/models"
 	"github.com/spf13/cobra"
 )
@@ -28,21 +29,47 @@ var installCmd = &cobra.Command{
 			return err
 		}
 
-		inputs, cv, err := prepareBlocknodeInputs(cmd, args)
+		shapeOverrides, err := common.ParseShapeOverrides(flagShape)
 		if err != nil {
 			return err
 		}
 
-		// Prompt for egress NIC and bandwidth independently of the host firewall —
-		// tc-egress traffic shaping applies regardless of whether the host
-		// firewall is enabled.
-		if err := common.ResolveEgressConfig(cmd, args, cv, &flagEgressInterface, &flagLinkRate); err != nil {
+		inputs, cv, err := prepareBlocknodeInputs(cmd, args)
+		if err != nil {
 			return err
+		}
+		// prepareBlocknodeInputs is shared with other block-node commands; patch
+		// in the install-only network field here.
+		inputs.Custom.ShapeOverrides = shapeOverrides
+
+		// The traffic-shaping gate is independent of the host firewall — it
+		// covers the BN workload network-policy plane and tc HTB shaping, not
+		// the host's own SSH/mgmt firewall.
+		trafficShapingEnabled, err := common.ResolveTrafficShapingConfig(cmd, args, cv)
+		if err != nil {
+			return err
+		}
+		inputs.Custom.TrafficShapingEnabled = trafficShapingEnabled
+
+		// Prompt for egress NIC and bandwidth, and resolve where the daemon
+		// binary comes from, only when traffic shaping is enabled — declining
+		// the gate above means there is nothing left to configure here.
+		var daemonSource workflowsteps.DaemonBinarySource
+		if trafficShapingEnabled {
+			if err := common.ResolveEgressConfig(cmd, args, cv, &flagEgressInterface, &flagLinkRate); err != nil {
+				return err
+			}
+			daemonSource, err = resolveDaemonBinarySource(cmd, args, inputs.Custom.Profile, cv)
+			if err != nil {
+				return err
+			}
 		}
 		// prepareBlocknodeInputs ran before the prompts; patch in the final values.
 		inputs.Custom.EgressInterface = flagEgressInterface
 		inputs.Custom.LinkRate = flagLinkRate
 
+		// Host firewall is independently gated from traffic shaping above — it
+		// applies regardless of whether the BN policy plane/tc shaping is enabled.
 		if err := common.ResolveHostFirewallConfig(cmd, args, cv); err != nil {
 			return err
 		}
@@ -82,6 +109,20 @@ var installCmd = &cobra.Command{
 
 		logx.As().Info().Msg("Successfully installed Hedera Block Node")
 
+		// Daemon activation is part of the traffic-shaping bundle, not a separate
+		// decision: a block node without its daemon has no ingress prioritization
+		// and nothing reconciling the daemon-owned nft sets from statusz, so
+		// enabling traffic shaping installs and provisions the daemon
+		// automatically (no separate prompt/flag). Skipped entirely when traffic
+		// shaping is disabled — there would be nothing for the daemon to do.
+		if trafficShapingEnabled {
+			if err := ensureBlockNodeDaemon(cmd, inputs.Custom.Namespace, daemonSource); err != nil {
+				return err
+			}
+		} else {
+			logx.As().Info().Msg("traffic shaping disabled (--traffic-shaping-enabled=false); skipping traffic-shaper daemon activation")
+		}
+
 		return nil
 	},
 }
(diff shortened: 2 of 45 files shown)


alex-au (approved):
Great work!

PR #605, opened 2026-05-22
feat(release): independent release pipelines for CLI and daemon binaries (#516, #517)

brunodam:
## Description

Completes epic #498 (Two-Binary Build Layout) by wiring the daemon into the top-level Taskfile build aggregator and splitting the release pipeline into two independent semantic-release workflows. CLI and daemon can now be released independently and at different cadences.

This pull request changes the following:

**#516 — Taskfile aggregator wiring**

* Top-level `task build` now produces both binaries (`build:cli:all` + `build:daemon:all`); `sources:`/`generates:` cover both.
* Per-binary `hash:cli:all` / `sign:cli:all` (and daemon equivalents) gain `deps: [build:*:all]` so they're self-sufficient — release configs invoke them directly.
* Vestigial top-level `hash:` and `sign:` aggregators removed (no remaining callers). `build:image:` retargeted to `deps: [build]`.

**#517 — Independent release processes**

* Two semantic-release configs at the repo root: `.releaserc_cli.json` (`tagFormat: solo-provisioner-v${version}`, writes `pkg/version/cli/VERSION`, 8 named assets) and `.releaserc_daemon.json` (`tagFormat: solo-provisioner-daemon-v${version}`, writes `pkg/version/daemon/VERSION`, 8 named assets).
* Two `workflow_dispatch`-only workflows: `flow-deploy-release-cli.yaml` and `flow-deploy-release-daemon.yaml`. The original single-pipeline `flow-deploy-release-artifact.yaml` and `.releaserc` are deleted.
* Both pipelines stay in the `0.x.y` range via the existing `releaseRules: [{ breaking: true, release: "minor" }]` override.

**`pkg/version` refactor**

* Split into symmetric `pkg/version/cli/` and `pkg/version/daemon/` subpackages — each embeds its own `VERSION` / `COMMIT` and exposes `Get` / `Cmd` / `Print`.
* Parent `pkg/version` keeps the shared `Info` struct + the `NewCmd(getter)` factory, plus a runtime-set `current Info` that subpackages register at `init()`. Shared code under `internal/...` continues to import `pkg/version` and read the running binary's version through `version.Get` / `Number` / `Commit` — no changes to call sites.

**Docs**

* Plan: [`docs/claude/plans/00516-taskfile-daemon-build-targets.md`](https://github.com/hashgraph/solo-weaver/blob/00516-taskfile-daemon-build-targets/docs/claude/plans/00516-taskfile-daemon-build-targets.md)
[...]

Issue #498, opened 2026-05-14
Epic A38_1 — Two-Binary Build Layout


leninmehedy:
Split the single binary into two separate compiled executables from the same repo.

PR #919, opened 2026-07-23
feat(catalog): add solo-provisioner-daemon entry with GPG-verified auto-download

### LINKED ISSUE #917: feat(catalog): add solo-provisioner-daemon entry so --daemon-version auto-download works
## Problem

`block node install` can auto-download the `solo-provisioner-daemon` binary from
the infrastructure catalog at a version selected by `--daemon-version` (defaulting
to the running CLI's own version). But `pkg/software/infrastructure-catalog.yaml`
has **no entry** for `solo-provisioner-daemon` — the catalog currently defines
only `cri-o`, `kubelet`, `kubeadm`, `kubectl`, `helm`, `k9s`, and friends.

As a result the auto-download path fails fast:

```
software 'solo-provisioner-daemon' not found in configuration
```

(the `softwareNotFoundErrorMsg` format string in `pkg/software/errors.go`). Any
install that reaches the daemon-activation path without `--daemon-bin` therefore
cannot obtain a daemon binary.

This is a pre-existing gap, not a regression: it was surfaced and documented as a
known follow-up while landing the install-time orchestration in #909, and is one
of the two reasons the traffic-shaping / daemon gate there defaults to opt-in
(`--traffic-shaping-enabled=false`) — a plain `block node install` would otherwise
fail by default on this missing catalog entry.

## Current workaround

Pass `--daemon-bin <path>` to bypass the catalog entirely (this is also the only
supported source for `--profile=local`, which has no downloadable release). The
resolution logic lives in `resolveDaemonBinarySource`
(`cmd/cli/commands/block/node/daemon_offer.go`): `--daemon-bin` skips the catalog,
`--daemon-version` selects the catalog version to download.

## Proposed fix

Add a `solo-provisioner-daemon` entry to `pkg/software/infrastructure-catalog.yaml`
following the shape of the existing single-binary entries (e.g. `kubectl` /
`kubeadm`): a download URL/filename template parameterised on `{{.OS}}`,
`{{.ARCH}}`, `{{.VERSION}}`, the extracted binary name, and a checksum so
`InstallDaemonBinaryStep` (`internal/workflows/steps/step_daemon.go`) can verify
the download.

Points to nail down while implementing:

- **Release artifact naming.** The daemon has its own release pipeline and tag
  namespace (`daemon-vX.Y.Z`, driven by `.releaserc_daemon.json` /
  `flow-deploy-release-daemon.yaml`), separate from the CLI's `vX.Y.Z`. The catalog
  entry's URL/version templating must resolve against the daemon tag namespace,
  not the CLI one.
- **Default version semantics.** `--daemon-version` defaults to the running CLI's
  own version. Confirm the catalog can resolve a daemon artifact for that version
  string, given the CLI and daemon version namespaces are independent and can
  drift — decide what happens when no daemon release matches the CLI version
  (fall back to a catalog default vs. fail with an actionable error).
- **Checksum management.** Match however the other catalog entries source and pin
  their checksums so the verify step in `InstallDaemonBinaryStep` stays meaningful.

## Acceptance

- [ ] `pkg/software/infrastructure-catalog.yaml` has a `solo-provisioner-daemon`
      entry resolving against the daemon (`daemon-vX.Y.Z`) release namespace.
- [ ] `block node install --traffic-shaping-enabled` **without** `--daemon-bin`
      (non-local profile) downloads and installs the daemon binary instead of
      failing with `software 'solo-provisioner-daemon' not found in configuration`.
- [ ] `--daemon-version <v>` selects the corresponding catalog version; the
      default (CLI's own version) resolves to a daemon artifact or fails with an
      actionable error.
- [ ] The downloaded binary passes the catalog-checksum verification in
      `InstallDaemonBinaryStep`.
- [ ] `--daemon-bin` continues to bypass the catalog unchanged.

### Related Issues

* Follow-up from #909 (install-time daemon orchestration; documented there as a
  known gap).

brunodam:
## Description

Adds a `solo-provisioner-daemon` entry to the infrastructure catalog so that
`block node install --traffic-shaping-enabled` (non-local profile, no
`--daemon-bin`) can auto-download and install the daemon, instead of failing with:

```
software 'solo-provisioner-daemon' not found in configuration
```

To make that auto-download resolve deterministically, this PR also **collapses the
two independent release pipelines back into a single pipeline that releases all
binaries together** (effectively reverting #605 / the epic #498 split, adapted to
today's tree — #614 brought it to `main`, #625 aligned tag namespaces, and #785
replaced `pkg/version` with `automa-saga/version`). CLI and daemon now ship in the
same GitHub release under one tag (`vX.Y.Z`), so `--daemon-version` (default: the
running CLI's own version) always resolves to a real daemon artifact — the
version-namespace drift the issue flagged goes away.

The daemon binary is built and released from this repo, so unlike the third-party
host tools its per-version SHA-256 **cannot be pinned** in the catalog: the catalog
is embedded in the very binary being released, which cannot know the checksum of
its own (or a later) artifact. Authenticity is therefore established by verifying
the release's detached **GPG signature** against the release signing key embedded
in the CLI (new `pkg/codesign`), which works for any released version.

### Release pipeline consolidation

* Replace `.releaserc_cli.json` + `.releaserc_daemon.json` with a single
  `.releaserc` (`tagFormat: v${version}`, all 16 CLI + daemon assets,
  `verifyRelease: task hash`, `prepare: task sign`; the daemon `--latest=false`
  hook is gone since there is now one release).
* Replace `flow-deploy-release-cli.yaml` + `flow-deploy-release-daemon.yaml` with a
  single `flow-deploy-release-artifact.yaml`.
* Add top-level `hash` / `sign` Taskfile aggregators that fan out to both binaries'
  `hash:{cli,daemon}:all` / `sign:{cli,daemon}:all`.

### Daemon catalog entry + signature verification

* New `pkg/codesign` package: embeds the release public key and verifies a detached
  `.asc` against it. It parses the key at the **packet level** because the published
  key is UID-stripped and `openpgp.ReadArmoredKeyRing` rejects a v4 entity with no
  identities; a signature is matched to a key by issuer key id and verified
  cryptographically.
* New `signedRelease` schema field on `ArtifactMetadata` for self-published,
  signature-verified artifacts that resolve for any version (no enumerated
  `versions` map, no pinned checksum).
* `daemonInstaller` owns a version-agnostic **download → signature-verify →
  install** path; an artifact that fails verification is deleted, never installed.
* `step_daemon.go`: single-namespace (`v`) release URL hint and corrected
  doc/comments (auto-download now verifies a signature, not a sha256).
* Flip the integration test that asserted the daemon must have **no** catalog entry
  → asserts it now resolves as a signed-release artifact.

Subkey-chain-to-offline-primary validation and key rotation/revocation are
deliberately **out of scope** — they belong to the self-upgrade work that owns the
fuller key model in `docs/dev/daemon/cli-verification-design.md`. The current
release key signs artifacts with its primary key directly, which this verifier
handles.

### Files changed

| File | Change |
|---|---|
| `.releaserc` (new), `.releaserc_cli.json` / `.releaserc_daemon.json` (deleted) | Single release config releasing all binaries under `vX.Y.Z` |
| `.github/workflows/flow-deploy-release-artifact.yaml` (new), `flow-deploy-release-{cli,daemon}.yaml` (deleted) | Single release workflow |
| `Taskfile.yaml` | Top-level `hash` / `sign` aggregators covering CLI + daemon |
| `pkg/codesign/codesign.go` (new) | Embedded-key detached-signature verifier (packet-level, UID-less-tolerant) |
| `pkg/codesign/keys/release-pubkey.asc` (new) | Embedded release public key + provenance `README.md` |
| `pkg/codesign/{codesign_test.go,testdata/}` (new) | Unit tests incl. verifying a real release signature |
| `pkg/software/config.go` | `SignedReleaseSpec` schema; `GetDefaultVersion`/`validate` tolerate signed-release entries |
| `pkg/software/infrastructure-catalog.yaml` | `solo-provisioner-daemon` signed-release entry |
| `pkg/software/daemon_installer.go` | Version-agnostic signed download/verify/install path |
| `pkg/software/config_it_test.go` | Flip negative daemon-entry test → positive; handle signed-release in catalog-shape tests |
| `internal/workflows/steps/step_daemon.go` | Single-namespace URL hint; doc/comment fixes |
| `CLAUDE.md`, `docs/dev/daemon/cli-verification-design.md` | Reflect single-pipeline release model |

### Review guide

**Verify the embedded key is authentic** (the one thing worth checking by hand).
The key is UID-less, so `gpg --show-keys` prints nothing — use `--list-packets`:

```bash
# Fingerprint of the embedded key (from its self-signature packet):
gpg --list-packets pkg/codesign/keys/release-pubkey.asc | grep -m1 'issuer fpr'
#   -> 9BBEC9EF1C3F21653824610BDB125DC2EB561F1C

# Same fingerprint on a real release signature (committed as testdata, no network):
gpg --list-packets pkg/codesign/testdata/release-asset.sha256.asc | grep -m1 'issuer fpr'
#   -> 9BBEC9EF1C3F21653824610BDB125DC2EB561F1C

# Authoritative end-to-end check (no gpg, no network) — embedded key verifies a
# real release signature:
go test ./pkg/codesign/ -run VerifiesRealReleaseSignature -v
```

The key is published (UID-stripped) at
`https://keys.openpgp.org/vks/v1/by-fingerprint/9BBEC9EF1C3F21653824610BDB125DC2EB561F1C`;
provenance is documented in `pkg/codesign/keys/README.md`.

**Checklist**

- [ ] `codesign.Verify` matches by issuer key id then verifies cryptographically;
      an untrusted-issuer or tampered artifact is rejected (`codesign_test.go`).
- [ ] Signed-release catalog entry has no pinned checksum and no `versions` map;
      `daemonInstaller` never installs an artifact that failed verification
      (`daemon_installer.go` deletes on verify failure).
- [ ] `.releaserc` lists all 16 assets; `task hash` / `task sign` produce both
      binaries' artifacts.
- [ ] `--daemon-bin` still bypasses the catalog unchanged.

**Test commands**

```bash
go test ./pkg/codesign/... ./pkg/software/...      # native; codesign incl. real-signature fixture
GOOS=linux GOARCH=amd64 go vet ./...
task lint                                          # 0 issues
task vm:test:unit                                  # Linux-only steps/common/bll (VM)
```

**Manual UAT (UTM VM)**

```bash
# Auto-download path: non-local, no --daemon-bin -> downloads + signature-verifies + installs
sudo solo-provisioner block node install --profile=testnet --traffic-shaping-enabled --non-interactive
systemctl is-active solo-provisioner-daemon        # active
# Negative: a tampered/unsigned binary at that URL must fail verification with an actionable error
```

**Risks / rollback**

* Release-time / catalog change only; no on-disk schema change. Revert the commit
  to restore the prior state. The `--daemon-bin` path is untouched.
* First real release after this merges must push a single `vX.Y.Z` tag; dry-run the
  `Deploy Release Artifact` workflow first to confirm it attaches all 16 assets.

### Related Issues

* Closes #917


alex-au (approved):
LGTM!

PR #921, opened 2026-07-23
fix(shaper): derive bn-backfill from outbound partner endpoints

### LINKED ISSUE #920: fix(daemon): derive bn-backfill from outbound partner endpoints — statusz has no peer_bn category
## Problem

The traffic-shaper reconciler maps statusz traffic categories to nft policy sets
in `internal/blocknode/shaper/policy_map.go`. That mapping was built against a
**provisional** statusz contract (the code itself flags it: *"The contract is
PROVISIONAL — confirming it with the BN team is a blocking dependency"*), and one
of its assumptions is wrong.

`policy_map.go` declares a literal outbound category `peer_bn` and binds it to the
compound `bn-backfill` set:

```go
CategoryPeerBN Category = "peer_bn"
// ...
CategoryPeerBN: {policyName: "bn-backfill", compound: true},
```

and `bucketizeEndpoints` populates `bn-backfill` only from outbound endpoints
whose `category == "peer_bn"`.

**But the statusz REST endpoints never emit `peer_bn`.** The only categories the
BN reports are:

- `partner`
- `publisher`
- `public`
- `restricted`

The peer-block-node / backfill membership is not its own category — it is
**derived** from the **outbound** endpoints reported under the `partner`
category (an *outbound* partner connection is a peer block node the BN backfills
from). So `bn-backfill` should be keyed on the compound address·port of the
outbound `partner` endpoints, not on a non-existent `peer_bn` category.

### Impact

Against a **real** BN, `peer_bn` never appears, so `bucketizeEndpoints`' outbound
loop matches nothing and `bn-backfill` **silently never populates** — the outbound
partner endpoints are dropped on the floor. This is masked today because:

- the `statuszmock` server echoes whatever category string the roster contains, so
  seeding a mock roster with `"category": "peer_bn"` makes the mock path work
  end-to-end against itself (this is what the #909 UAT step 12 and the #916 UAT
  exercise) — it validates the wiring, not the real category contract;
- the daemon poll loop (#914 / #916) only recently began driving reconciliation
  at all, so the gap had no runtime surface until now.

### Direction-awareness

A consequence worth calling out: the same `partner` category string now maps to
**two different sets depending on direction** — inbound `partner` and outbound
`partner` are not the same policy. `bucketizeEndpoints` already reads the inbound
and outbound endpoint lists separately, so the fix keys the binding on
(direction, category), not category alone.

## Proposed fix

- Replace the literal `peer_bn` category with a direction-aware binding: the
  compound `bn-backfill` set is fed by **outbound** endpoints of category
  `partner`; the inbound `partner` endpoints continue to feed their existing set.
- Add the `public` category to the known vocabulary and confirm it stays
  **unmapped** — `bn-public-out` is a `--from-entity world` port match, not a
  statusz-reconciled set (per #909), so `public` should be recognized-but-ignored
  rather than silently unknown.
- Update `statuszmock` so its roster/served categories reflect the real contract
  (`partner`/`publisher`/`public`/`restricted`, with the outbound peers under
  `partner`), instead of the fictional `peer_bn`.
- Update the reconciler tests and the `policy_map.go` doc comments; drop the
  "PROVISIONAL" caveat once the BN team confirms the vocabulary.
- Confirm the exact inbound-vs-outbound split for each set with the BN team
  before finalizing (which of `partner`/`publisher`/`restricted` appear on
  inbound vs outbound, and how each maps to `bn-publisher` / `bn-partner-out` /
  `bn-restricted` / `bn-backfill`).

## Acceptance

- [ ] The reconciler derives `bn-backfill` from outbound `partner` endpoints
      (compound address·port); no code path references a `peer_bn` category.
- [ ] Against a statusz source emitting only `partner`/`publisher`/`public`/
      `restricted`, all daemon-owned sets — including `bn-backfill` — reconcile
      correctly.
- [ ] `public` is recognized and deliberately left unmapped (no "unknown
      category" surprise; `bn-public-out` stays port-match-driven).
- [ ] `statuszmock` serves the real category vocabulary; the UAT rosters no longer
      rely on `peer_bn`.
- [ ] The inbound/outbound → policy-set mapping is confirmed with the BN team and
      the PROVISIONAL caveat is removed from `policy_map.go`.

### Related Issues

* Surfaced during review of #916 (daemon statusz poll scheduler); the scheduler
  wiring is correct — this is a pre-existing category-contract bug in the
  reconciler stories, independent of #916.

brunodam:
## Description

The traffic-shaper reconciler mapped statusz traffic categories to nft policy
sets against a **provisional** contract, and one assumption was wrong: it keyed
the compound `bn-backfill` set on a literal `peer_bn` category. A real block node
never emits `peer_bn` — its only categories are `partner`, `publisher`, `public`,
and `restricted`, and the peer block nodes it backfills from are reported as
**outbound `partner`** connections. So against a real BN the outbound bucketize
loop matched nothing and `bn-backfill` **silently never populated**. The gap was
masked because `statuszmock` echoes whatever category the roster carries, so
`peer_bn` rosters validated the wiring against themselves, not the real contract.

The fix makes the binding table **direction-aware**. The same `partner` string
now maps to a different set depending on direction — inbound `partner` →
`bn-partner-out`, outbound `partner` → the compound `bn-backfill` — so the table
is keyed on `(direction, category)` rather than category alone. `peer_bn` is
removed entirely, and `public` is added as a recognized-but-unmapped category
(`bn-public-out` stays a port-match set, not statusz-reconciled), so a `public`
endpoint is deliberately dropped rather than treated as an unknown category.

The category vocabulary is confirmed with the BN team, so the `PROVISIONAL`
caveat is dropped from `policy_map.go` and `statuszmock`.

### Files changed

| File | Change |
|---|---|
| `internal/blocknode/shaper/policy_map.go` | `Direction` type + composite `(direction, category)` binding key; drop `peer_bn`, add `public`; `bn-backfill` ← outbound `partner`; drop PROVISIONAL caveat |
| `internal/blocknode/shaper/reconciler.go` | `bucketizeEndpoints` matches inbound/outbound endpoints against direction-specific bindings (new `bucketize` helper) |
| `internal/blocknode/shaper/policy_map_test.go` | Composite-key fixtures; new direction-split and public-unmapped cases |
| `internal/blocknode/shaper/reconciler_test.go` | Composite-key fixtures; outbound `partner` replaces `peer_bn` |
| `internal/blocknode/shaper/statusz_client_test.go`, `statusz_client_mock_test.go` | Wire samples use outbound `partner` |
| `internal/daemon/blocknode/statuszmock/server.go`, `server_test.go`, `cmd/main.go` | Serve/document the real vocabulary; drop PROVISIONAL caveat |
| `docs/dev/daemon/traffic-shaper-statusz.md`, `docs/dev/traffic-shaper-demo.md` | Mapping tables reflect the direction-aware binding; fix stale `bn-partner` → `bn-partner-out` |

### Scope boundary

Minimal mapping change: only the `partner` split is confirmed and encoded.
`publisher`/`restricted` stay inbound (as before); if the BN's real split for
those differs, the composite-key structure already supports a follow-up without
further refactor.

### Review guide

**Checklist**
- `policy_map.go` — `categoryBindings` has exactly the four `(direction, category)`
  keys; only `{Outbound, partner}` is `compound: true`. No `peer_bn` anywhere.
- `reconciler.go` `bucketize` — inbound endpoints look up `{Inbound, cat}`,
  outbound look up `{Outbound, cat}`; wildcard/empty ports still skipped for the
  compound set.
- `policy_map_test.go:TestComputePolicyDeltas_PartnerSplitsByDirection` — proves
  inbound and outbound `partner` produce `bn-partner-out` and `bn-backfill`
  separately.
- `policy_map_test.go:TestComputePolicyDeltas_PublicRecognizedButUnmapped` —
  `public` yields no delta and no live-set read.
- `TestCategoryBindings_PolicyNamesAreCanonical` still guards the policy-name set.

**Test commands**
```bash
go test -tags='!integration' ./internal/blocknode/shaper/... ./internal/daemon/blocknode/statuszmock/...
task lint
```

**Manual UAT** — serve a real-vocabulary roster and confirm `bn-backfill`
populates from outbound `partner`:
```bash
cat > roster.json <<'JSON'
{
  "inbound":  [ { "remote": {"address": "10.20.1.0/24", "port": "*"}, "category": "partner" } ],
  "outbound": [ { "remote": {"address": "10.30.5.7", "port": "43473"}, "category": "partner" } ]
}
JSON
go run ./internal/daemon/blocknode/statuszmock/cmd --addr 127.0.0.1:8080 --roster roster.json &
solo-provisioner block node reconcile-shaper --statusz-url http://127.0.0.1:8080 --check --output json
# expect desired["bn-backfill"] == ["10.30.5.7 . 43473"] and desired["bn-partner-out"] == ["10.20.1.0/24"]
```

**Risks / rollback** — re-keying `CategoryEndpoints` touches every reader in the
package; a missed call site is a compile error, not a silent bug. Rollback:
revert this PR (the inert `peer_bn` binding returns).

### Related Issues

* Closes #920

### DIFF
--- a/docs/dev/daemon/traffic-shaper-statusz.md
+++ b/docs/dev/daemon/traffic-shaper-statusz.md
@@ -66,7 +66,7 @@ Both endpoints are REST/JSON, resolved relative to the configured base URL:
 
 - `remote.address` is the source (inbound) or destination (outbound) host or
   CIDR. It is the value written into the nft set.
-- `remote.port` is only meaningful for the outbound `peer_bn` category, where
+- `remote.port` is only meaningful for the outbound `partner` category, where
   the backfill set is keyed on `address . port`. Inbound categories ignore the
   port (it is the BN's listener port, not part of the source allowlist).
 - Fields the shaper does not consume (`scheme`, `protocol`, `certificate`) are
@@ -83,14 +83,21 @@ a shared config file.
 | statusz endpoint | category | policy / nft set | key shape |
 |---|---|---|---|
 | inbound-clients | `publisher` | `bn-publisher` | ipv4 host/CIDR |
-| inbound-clients | `partner` | `bn-partner` | ipv4 host/CIDR |
+| inbound-clients | `partner` | `bn-partner-out` | ipv4 host/CIDR |
 | inbound-clients | `restricted` | `bn-restricted` | ipv4 host/CIDR |
-| outbound-clients | `peer_bn` | `bn-backfill` | compound `ip . port` |
+| outbound-clients | `partner` | `bn-backfill` | compound `ip . port` |
 
-Categories not in this table (for example `public`, or the operator-curated
-management set) are **never touched** by the monitor. A `public` source is
-expressed as the absence of a source-match on its rule, not as a set element, so
-it is intentionally dropped during bucketing.
+The mapping is keyed on **(direction, category)**, not category alone: the same
+`partner` string maps to `bn-partner-out` inbound and to the compound
+`bn-backfill` outbound. A peer block node this BN backfills from is reported as
+an outbound `partner` connection — there is no distinct `peer_bn` category.
+
+The `public` category is **recognized but deliberately unmapped**: a `public`
+source is expressed as the absence of a source-match on its rule (the
+`bn-public-out` port-match set), not as a statusz-reconciled set element, so it
+is intentionally dropped during bucketing rather than treated as an unknown
+category. The operator-curated management sets are likewise never touched by the
+monitor.
 
 Each owned category is reconciled on every successful poll: an address that
 drops out of statusz is removed from its set, not left stale. A poll that
@@ -160,7 +167,7 @@ Roster file shape:
     { "remote": {"address": "10.20.1.0/24", "port": "*"}, "category": "partner" }
   ],
   "outbound": [
-    { "remote": {"address": "10.30.5.7", "port": "43473"}, "category": "peer_bn" }
+    { "remote": {"address": "10.30.5.7", "port": "43473"}, "category": "partner" }
   ]
 }
 ```
--- a/docs/dev/traffic-shaper-demo.md
+++ b/docs/dev/traffic-shaper-demo.md
@@ -42,24 +42,28 @@ exactly, and each `--stamp` references one of the fixed class names (`publisher`
 creates the `inet weaver` table:
 
 ```bash
-sudo solo-provisioner network policy create --name bn-publisher  --stamp publisher --ports 40840
-sudo solo-provisioner network policy create --name bn-partner    --stamp partner   --ports 40980,40981
-sudo solo-provisioner network policy create --name bn-restricted --deny
+sudo solo-provisioner network policy create --name bn-publisher   --stamp publisher --ports 40840
+sudo solo-provisioner network policy create --name bn-partner-out --stamp partner   --ports 40980,40981
+sudo solo-provisioner network policy create --name bn-restricted  --deny
 sudo solo-provisioner network policy create --name bn-backfill    --stamp reserve-egress --reply-stamp backfill-response
 ```
 
-| category (statusz) | policy / nft set | create flags |
+| statusz (direction, category) | policy / nft set | create flags |
 |---|---|---|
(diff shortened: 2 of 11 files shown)


alex-au (approved):
LGTM!

PR #922, opened 2026-07-23
feat(blocknode): make helm install/upgrade timeout configurable via --timeout

### LINKED ISSUE #912: Make the helm install/upgrade timeout configurable (currently hard-coded 5m)
## Summary

`block node install` / `upgrade` run `helm install/upgrade` with `--atomic` and a **hard-coded
5-minute timeout**. The timeout is not exposed as a CLI flag or config field, so an operation whose
first boot legitimately takes longer than 5 minutes is rolled back even though it would otherwise
succeed.

## Where (observed on v0.24.0)

- `pkg/helm/interface.go` — `DefaultTimeout = 5 * time.Minute`.
- `pkg/helm/options.go` — `InstallChartOptions` / `UpgradeChartOptions` already carry a
  `Timeout time.Duration` field, and `pkg/helm/manager.go` honours it if set…
- `internal/blocknode/chart.go` — …but `InstallChart()` / `UpgradeChart()` **hard-code**
  `Timeout: helm.DefaultTimeout` at the call site. There is no `--timeout` flag or config field to
  override it.

## Impact

The block-node Helm chart's default plugin-delivery path is a `resolve-plugins` init container that
resolves plugins from Maven at startup (~4 minutes cold on a fresh/changed plugins PVC). That cold
resolve races the 5-minute `--atomic` timeout and can roll back an otherwise-successful
install/upgrade.

Downstream (the hiero-block-node `e2e-node-operator` lifecycle test) has to avoid the default flow
entirely — using a plugins-baked `-solo-dev` image with `plugins.names: ""` to skip `resolve-plugins`
— largely to stay under this timeout. A configurable timeout would let that test (and real operators
with slow plugin resolution or large state loads) use the chart's default plugin flow.

## Request

Expose the install/upgrade timeout as a `--timeout` CLI flag and/or a config field, plumbed into
`InstallChartOptions.Timeout` / `UpgradeChartOptions.Timeout` at the `internal/blocknode/chart.go`
call sites. Keep `5m` as the default.

## Suggested triage

`Feature Enhancement`, `BN`, `P3` (labels not set here — filer has read-only access).

alex-au:
## Summary

`block node install` / `upgrade` ran Helm with `--atomic` and a **hard-coded 5-minute
timeout** that no flag or config field could override. An install/upgrade whose first
boot legitimately takes longer than 5 minutes — most commonly the chart's default
`resolve-plugins` init container resolving plugins from Maven (~4 min cold on a fresh or
changed plugins PVC) — raced that budget and got rolled back even though it would have
succeeded.

This PR exposes the timeout as a `--timeout` duration flag on `block node install` and
`block node upgrade`, plumbed through `models.BlockNodeInputs` into the existing
`InstallChartOptions.Timeout` / `UpgradeChartOptions.Timeout` fields (which the Helm
manager already honours). The default stays **5m**, and any code path that leaves the
timeout unset falls back to `helm.DefaultTimeout`, so no existing behaviour changes.

The flag accepts a Go duration string (e.g. `10m`, `600s`, `1h`) — matching upstream
Helm's own `--timeout` — not a bare seconds integer.

### Changes

| File | Change |
|---|---|
| `pkg/models/inputs.go` | Add `Timeout time.Duration` to `BlockNodeInputs`; `Validate()` rejects negative values |
| `cmd/cli/commands/common/flags_blocknode.go` | New `FlagHelmTimeout()` factory — `--timeout`, default `helm.DefaultTimeout` |
| `cmd/cli/commands/block/node/node.go` | Declare shared `flagHelmTimeout` var |
| `cmd/cli/commands/block/node/install.go`, `upgrade.go` | Register `--timeout` on both commands |
| `cmd/cli/commands/block/node/init.go` | Wire flag into `inputs.Custom.Timeout` |
| `internal/blocknode/chart.go` | `InstallChart`/`UpgradeChart` use `resolveHelmTimeout()` (configured value; `<= 0` → `helm.DefaultTimeout`) |
| `docs/quickstart.md` | Add `--timeout` to install + upgrade flag tables |
| `internal/blocknode/chart_timeout_test.go` | New: timeout propagation + zero/negative fallback via helm mock |
| `pkg/models/validation_test.go` | Timeout validation cases |

### Scope boundary

Flag only (no Viper config key). `--timeout` is registered on `install` and `upgrade`
only — `reconfigure`/`reset`/`check`/`uninstall` and the other chart installers
(alloy, MetalLB, teleport, ESO, metrics-server, prometheus-operator-crds) are unchanged
and keep the 5m default.

### Unit

```bash
# timeout propagation + zero/negative fallback (helm mock)
go test -race -tags='!integration' ./internal/blocknode/... -run 'Timeout' -v

# BlockNodeInputs.Validate rejects negative, accepts zero/positive
go test -race -tags='!integration' ./pkg/models/... -run 'TimeoutValidation' -v
```

Full unit suite (run in the UTM VM — several packages are Linux-only):

```bash
task vm:test:unit
```

### Manual UAT

> Run on a host/VM with a Kubernetes cluster provisioned by this tool
> (`sudo solo-provisioner kube cluster install ...` first if needed).

**UAT-1 — `--timeout` appears in help with the 5m default**

```bash
solo-provisioner block node install --help
solo-provisioner block node upgrade --help
```

*Expected:* both list `--timeout duration` with description ending
`... (default 5m0s)`.

**UAT-2 — duration format is accepted; bare seconds is rejected**

```bash
# valid Go durations parse
solo-provisioner block node install --profile=local --timeout=10m ... --help   # no parse error
# a bare integer is rejected (Go duration requires a unit)
solo-provisioner block node install --profile=local --timeout=600 --help
```

*Expected:* `--timeout=10m` parses fine; `--timeout=600` fails with
`invalid argument "600" for "--timeout" flag: time: missing unit in duration "600"`.
(`600s`, `1h`, `90m` are all valid.)

**UAT-3 — negative timeout is rejected by input validation**

```bash
solo-provisioner block node install --profile=local --timeout=-5m ...
```

*Expected:* fails fast with `invalid user inputs: invalid timeout: -5m0s (must not be negative)`
before any Helm operation runs.

**UAT-4 — a raised timeout lets a slow first-boot install succeed (the reported case)**

Use the chart's default plugin flow (do **not** pass `plugins.names: ""`), so the
`resolve-plugins` init container performs a cold Maven resolve:

```bash
sudo solo-provisioner block node install \
  --profile=local \
  --plugin-preset=tier1-lfh \
  --timeout=15m \
  --daemon-bin=/path/to/solo-provisioner-daemon
```

*Expected:* the install waits up to 15m for readiness and completes successfully; it is
**not** rolled back at the 5-minute mark. Watch progress with:

```bash
kubectl -n block-node get pods -w
kubectl -n block-node logs <bn-pod> -c resolve-plugins
```

**UAT-5 — default (flag omitted) is still 5m**

```bash
sudo solo-provisioner block node install --profile=local ... --daemon-bin=...
```

*Expected:* behaves exactly as before this PR — the Helm operation uses a 5-minute
`--atomic` budget. Confirm in the CLI logs (`solo-provisioner` log) that the helm
install/upgrade timeout is `5m0s`.

**UAT-6 — upgrade honours `--timeout` too**

```bash
sudo solo-provisioner block node upgrade --profile=local --timeout=12m --values=/path/to/values.yaml
```

*Expected:* the upgrade uses a 12-minute budget; omitting `--timeout` keeps 5m.

## Risks / rollback

- Low risk: the Helm layer already honoured `Timeout`; this stops hard-coding it at the
  block-node call site. The `<= 0` fallback preserves today's 5m behaviour for every path
  that does not set the new field (migrations, tests, non-install/upgrade commands).
- CLI-surface change → `docs/quickstart.md` updated in the same PR.
- No state/schema/template changes. Rollback = revert the commit.

Closes #912

### DIFF
--- a/cmd/cli/commands/block/node/init.go
+++ b/cmd/cli/commands/block/node/init.go
@@ -311,6 +311,7 @@ func prepareBlocknodeInputs(cmd *cobra.Command, args []string) (*models.UserInpu
 			PluginList:          pluginList,
 			EgressInterface:     flagEgressInterface,
 			LinkRate:            flagLinkRate,
+			Timeout:             flagHelmTimeout,
 		},
 	}
 
--- a/cmd/cli/commands/block/node/install.go
+++ b/cmd/cli/commands/block/node/install.go
@@ -130,6 +130,7 @@ var installCmd = &cobra.Command{
 func init() {
 	installCmd.Flags().StringVar(&flagChartVersion, "chart-version", "", "Helm chart version to use")
 	common.FlagValuesFile().SetVarP(installCmd, &flagValuesFile, false)
+	common.FlagHelmTimeout().SetVarP(installCmd, &flagHelmTimeout, false)
 	common.RegisterHostFirewallFlags(installCmd)
 	common.RegisterTrafficShapingFlags(installCmd)
 	common.RegisterEgressFlags(installCmd, &flagEgressInterface, &flagLinkRate)
--- a/cmd/cli/commands/block/node/node.go
+++ b/cmd/cli/commands/block/node/node.go
@@ -3,6 +3,8 @@
 package node
 
 import (
+	"time"
+
 	"github.com/hashgraph/solo-weaver/cmd/cli/commands/common"
 	"github.com/hashgraph/solo-weaver/pkg/models"
 	"github.com/spf13/cobra"
@@ -43,6 +45,7 @@ var (
 	flagShape                []string
 	flagDaemonBin            string
 	flagDaemonVersion        string
+	flagHelmTimeout          time.Duration
 
 	nodeCmd = &cobra.Command{
 		Use:   "node",
--- a/cmd/cli/commands/block/node/upgrade.go
+++ b/cmd/cli/commands/block/node/upgrade.go
@@ -72,4 +72,5 @@ func init() {
 	common.FlagWithStorageReset().SetVarP(upgradeCmd, &flagWithReset, false)
 	common.FlagValuesFile().SetVarP(upgradeCmd, &flagValuesFile, false)
 	common.FlagNoReuseValues().SetVarP(upgradeCmd, &flagNoReuseValues, false)
+	common.FlagHelmTimeout().SetVarP(upgradeCmd, &flagHelmTimeout, false)
 }
--- a/cmd/cli/commands/common/flags_blocknode.go
+++ b/cmd/cli/commands/common/flags_blocknode.go
@@ -2,6 +2,12 @@
 
 package common
 
+import (
+	"time"
+
+	"github.com/hashgraph/solo-weaver/pkg/helm"
+)
+
 // Flag descriptor factories for the `block node` command tree.
 //
 // Each factory returns a fresh FlagDefinition value so callers always receive an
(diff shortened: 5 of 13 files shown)


brunodam (approved):
LGTM

Issue #381, opened 2026-02-26
Upgrade fails when per-pod PVCs remain Pending (block-node StatefulSet) — upgrade/installation rollback


mahmoudian1:
### Bill of Materials, Application, or CLI Version

sudo solo-provisioner block node install --profile=local --chart-version=0.26.2 --values=/path/to/values-override.yaml

### Describe the bug

When running solo-provisioner to install/upgrade the block node (chart v0.26.2), the StatefulSet creates per-pod PVCs (e.g. live-storage-block-node-block-node-server-0) that remain Pending because no matching PVs exist. The Pending PVCs prevent pod scheduling, Helm times out waiting for the pod to become Ready, and because Helm runs with atomic=true the release is rolled back and the upgrade fails.

This issue groups the relevant troubleshooting logs and requests clearer documentation and/or a migration approach so upgrades/installations do not fail in this way.

### Describe the expected behavior

This issue shouldn't happen and the dependent resources should be created. The installer/upgrade should detect missing per-pod PVs and emit a clear, actionable error pointing to the Pending per-pod PVC names and recommended remediation steps.

### To Reproduce

Prepare a host machine and run:

```
sudo solo-provisioner block node install --profile=local --chart-version=0.26.2 --values=/path/to/values-override.yaml
(Or run upgrade similarly:)
```
```
sudo solo-provisioner block node upgrade --profile=local --chart-version=0.26.2 --with-reset --values=values-override.yaml
Watch the block-node namespace resources:
```
```
kubectl -n block-node get pods
kubectl -n block-node get pvc
kubectl -n block-node get pv
```
Observed behavior (logs / excerpts)
Pod pending due to unbound PVCs:
```
weaver@bn-lfh01-previewnet:~$ kubectl -n block-node get po
NAME                             READY   STATUS    RESTARTS   AGE
block-node-block-node-server-0   0/1     Pending   0          11s
Pod describe shows Pending + unbound PVCs:
```
```
Events:
  Type     Reason            Age   From               Message
  ----     ------            ----  ----               -------
  Warning  FailedScheduling  22s   default-scheduler  0/1 nodes are available: pod has unbound immediate PersistentVolumeClaims.
PVCs show Pending (per-pod PVCs), while other PV/PVC entries show some PVs already Bound:
```

```
kubectl -n block-node get pvc
NAME                                                  STATUS    VOLUME                    CAPACITY
archive-storage-block-node-block-node-server-0        Pending
[...]

PR #931, opened 2026-07-28
feat(block-node): skip managed plugins PVC for plugins-baked images

### LINKED ISSUE #913: Allow opting out of the managed plugins PVC (or support an ephemeral plugins volume) for plugins-baked images
## Summary

The provisioner **always** provisions a plugins PVC and injects it into the block-node chart as
`persistence.plugins.existingClaim`; the chart then mounts it **read-only over the image's baked-in
plugins dir** (`/opt/hiero/block-node/app-<ver>/plugins`). There is no way to tell the provisioner
"this image already ships its plugins — don't provision or mount a plugins PVC." For a plugins-baked
image that managed PVC is counterproductive.

## Where (observed on v0.24.0)

- `internal/blocknode/optional_storage.go` — the `plugins` storage is registered unconditionally
  (`MinVersion: 0.28.1`, no max version); there is no `managePlugins`/`storage.plugins: none` switch.
- `internal/blocknode/values.go` — `injectPersistenceOverrides()` always emits
  `persistence.plugins.existingClaim: plugins-storage-pvc`.
- `internal/templates/files/block-node/{full,nano}-values.yaml` — the plugins volume is always
  injected.

## Impact (plugins-baked image, e.g. the `-solo-dev` variant with `plugins.names: ""`)

1. **The read-only PVC mount shadows the image's baked plugins** → the node comes up with an empty
   plugins dir → `No BlockMessagingFacility provided`.
2. Being a **persistent** PVC, it **carries stale jars across an upgrade** (the previous version's
   jars remain). The BN loads *every* jar in the plugins dir, so a stale `verification-<old>.jar`
   loaded alongside the new one causes `BAD_BLOCK_PROOF`; neither Helm nor the provisioner clears it.

Downstream (the hiero-block-node `e2e-node-operator` test) works around this with a `seed-plugins`
init container that (a) clears the forced PVC, (b) copies the image's baked plugins into it, and (c)
prunes to the selected profile — steps that exist solely to compensate for the forced, persistent PVC.

## Request

Either of these would resolve it:

- **(a) No managed plugins volume** — don't inject `persistence.plugins.existingClaim` (the chart
  already gates the plugins volume + mount on that claim), so the main container reads the image's
  baked `app-<ver>/plugins` directly. Simplest; suits per-image plugin sets.
- **(b) Ephemeral (`emptyDir`) plugins volume** — provision the plugins volume as a per-pod
  `emptyDir` instead of a persistent PVC. An init container can still (re)populate it each boot, and
  because it starts empty there is no cross-upgrade staleness. Preserves runtime plugin selection.

Either removes the downstream clear/copy workaround and matches how a plugins-baked image is meant to
run.

## Suggested triage

`Feature Enhancement`, `BN` (labels not set here — filer has read-only access).

alex-au:
## Summary

For a plugins-baked block-node image (e.g. the `-solo-dev` variant, `plugins.names: ""`), the
provisioner still provisioned a managed `plugins-storage` PVC and forced
`blockNode.persistence.plugins.existingClaim` into the chart values. The chart then mounted that empty
PVC **read-only over the image's baked** `app-<ver>/plugins` directory (→ `No BlockMessagingFacility
provided`), and because the PVC is persistent it carried **stale jars across upgrades**.

This PR makes the provisioner **mirror the chart's own plugins-volume condition**: it provisions the
plugins PV/PVC and injects `existingClaim` **iff the effective `plugins.names` is non-empty**. When
empty, it provisions nothing, injects no `existingClaim`, and emits an explicit `plugins.names: ""` so
Helm does not re-apply the chart's own non-empty default — the chart then renders no plugins volume and
the container reads the image's baked plugins directly. No new CLI flag; the trigger is the plugins
signal the provisioner already resolves (`--plugins`/`--plugin-preset`, else the `--values`
`plugins.names`, else the non-empty default).

Closes #913.

## Problem

The `plugins` optional storage was gated only on chart version (`>=0.28.1`), independent of whether any
plugins are actually being managed. The block-node chart, however, already distinguishes the modes
(`charts/block-node-server/templates/statefulset.yaml`):

| Chart element | Condition |
|---|---|
| pod volume (existingClaim ref) | `if persistence.plugins.existingClaim` |
| main-container read-only mount over `app-<ver>/plugins` | `if (plugins.names OR existingClaim)` |
| chart's own volumeClaimTemplate | `if (plugins.names AND NOT existingClaim)` |
| Maven `resolve-plugins` init | `if plugins.names` |

So with `plugins.names` empty **and** no `existingClaim`, the chart renders **no** plugins volume, mount,
VCT, or Maven init — the baked plugins are used. The only thing defeating that was the provisioner's
unconditional `existingClaim` injection. The volumeClaimTemplate cannot fire without a non-empty
`plugins.names`, so the `Pending`-PVC failure that originally justified the forced PVC (#381) **cannot
recur** in the baked case.

## Solution

A single predicate, `Manager.managesPluginsStorage()` (inverse of the pure
`EffectivePluginsNamesEmpty(pluginList, valuesFile)`), gates the five plugins consumers:

- `renderDefaultValues` — `IncludePlugins` (base-template volume/mount) only when managed.
- `injectPersistenceOverrides` — skip the plugins `existingClaim` entry when not managed.
- `injectPluginsConfig` — emit an explicit `plugins.names: ""` in the baked case.
- `CreatePersistentVolumes` — skip the plugins PV/PVC and its PVC-bind wait.
- `StorageMigration.Applies` — the plugins upgrade migration does not apply in the baked case (signal
  passed via a context bool set before migration filtering).

`verification` and `application-state` storage are untouched. `emptyDir` (issue option b) is intentionally
not implemented — the chart has no emptyDir source for the plugins volume (only `existingClaim` or its own
VCT), so it would require an upstream chart change; option (a) is implemented here.

### Tier 0 — unit (no cluster; the core logic)

```bash
go test -tags='!integration' ./internal/blocknode/ -v -run \
  'EffectivePluginsNamesEmpty|InjectPersistence|InjectPluginsConfig|PluginsBakedTwoLayerMerge|PluginsStorageMigration_Applies_BakedSignal'
# full unit suite (bll/blocknode pulls in linux-only internal/mount):
task vm:test:unit
```

`TestPluginsBakedTwoLayerMerge_ChartSeesEmpty` is the load-bearing one: it runs the real value pipeline
then simulates Helm coalescing over the chart's non-empty default and asserts the chart ends up with an
empty `plugins.names` in the baked case (non-empty otherwise).

### Tier 1 — local dev UAT (executable)

**Pinned versions:** chart `0.39.0` (>= 0.37.0, so `application-state` is active and `verification` is
retired), baked image `ghcr.io/hiero-ledger/hiero-block-node-solo-dev:0.39.0`.

**Prerequisites.** A Linux host **with systemd** where the provisioner has installed a cluster
(`solo-provisioner kube cluster install --profile local`). The UTM VM (`task vm:start`) is the standard
environment and is what these steps assume. Note: bare `kind` on Docker is *not* suitable yet — the
provisioner installs a tc-egress systemd service that fails on systemd-less nodes (#915).

**The observable outcome does not need a healthy pod.** The PVC/PV set and the rendered StatefulSet spec
land the moment Helm applies, so the assertions below hold even if the BN pod stays `Pending`/`CrashLoop`
under local resource limits.

#### Case 1 — baked image uses its own plugins (the fix)

Save `baked-values.yaml`:

```yaml
# baked-values.yaml — solo-dev ships ALL plugins in the image; the empty
# plugins.names is the signal to skip the managed plugins PVC (#913).
image:
  repository: ghcr.io/hiero-ledger/hiero-block-node-solo-dev
  tag: "0.39.0"

plugins:
  names: ""

resources:
  requests:
    cpu: "500m"
    memory: "1Gi"
  limits:
    cpu: "1"
    memory: "1.5Gi"

blockNode:
  config:
    JAVA_OPTS: '-Xms1G -Xmx1G'
  logs:
    level: "INFO"

kubepromstack:
  enabled: false
loki:
  enabled: false
promtail:
  enabled: false
```

Install:

```bash
sudo solo-provisioner block node install \
  --profile local \
  --chart-version 0.39.0 \
  --base-path /opt/weaver/block-node-storage \
  --values baked-values.yaml \
  --load-balancer-enabled=false \
  --skip-hardware-checks
```

Assert (each line prints `PASS: ...` on success):

```bash
kubectl -n block-node get pvc -o name | grep -q plugins && echo 'FAIL: plugins PVC exists' || echo 'PASS: no plugins PVC'
kubectl -n block-node get pv  -o name | grep -q plugins-storage && echo 'FAIL: plugins PV exists' || echo 'PASS: no plugins PV'
kubectl -n block-node get statefulset -o yaml | grep -q 'plugins-storage' && echo 'FAIL: plugins volume/mount rendered' || echo 'PASS: no plugins volume/mount'
kubectl -n block-node get statefulset -o yaml | grep -Eq 'app-[0-9.]+/plugins' && echo 'FAIL: baked-plugins shadow mount present' || echo 'PASS: no shadow mount over baked plugins'
```

Expected `kubectl -n block-node get pvc` (note: **no** `plugins-storage-pvc`; `application-state` present at 0.39.0):

```
NAME                            STATUS   VOLUME                          ...
archive-storage-pvc             Bound    archive-storage-pv
live-storage-pvc                Bound    live-storage-pv
logging-storage-pvc             Bound    logging-storage-pv
application-state-storage-pvc   Bound    application-state-storage-pv
```

Optional (Tier 2, needs the pod to start): `kubectl -n block-node logs sts/block-node-block-node-server`
contains no `No BlockMessagingFacility provided`.

#### Case 2 — managed/download mode unchanged (regression)

Save `managed-values.yaml` (plain image, chart-default `plugins.names` → Maven download → managed PVC):

```yaml
# managed-values.yaml — no plugins.names override; the provisioner manages the plugins PVC as before.
resources:
  requests:
    cpu: "500m"
    memory: "1Gi"
  limits:
    cpu: "1"
    memory: "1.5Gi"

kubepromstack:
  enabled: false
loki:
  enabled: false
promtail:
  enabled: false
```

```bash
sudo solo-provisioner block node install \
  --profile local \
  --chart-version 0.39.0 \
  --base-path /opt/weaver/block-node-storage \
  --values managed-values.yaml \
  --load-balancer-enabled=false \
  --skip-hardware-checks

# Assert the managed PVC is back exactly as before:
kubectl -n block-node get pvc plugins-storage-pvc -o jsonpath='{.status.phase}' | grep -q Bound \
  && echo 'PASS: plugins-storage-pvc Bound (managed mode unchanged)' || echo 'FAIL'
```

> This path is unchanged by the PR and includes the ~4-min Maven resolve. If resources/time are tight,
> Case 2 is safe to **skip** — it is covered by `TestInjectPersistenceOverrides_ManagedKeepsPluginsClaim`
> and `TestPluginsStorageMigration_Applies_BakedSignal` (managed sub-case).

#### Case 3 — upgrade managed → baked (optional; also unit-covered)

After a Case-2 install:

```bash
sudo solo-provisioner block node upgrade \
  --profile local --chart-version 0.39.0 \
  --base-path /opt/weaver/block-node-storage \
  --values baked-values.yaml --load-balancer-enabled=false --skip-hardware-checks
```

Expected: no new plugins PV/PVC provisioned; the old `plugins-storage-pvc`/`-pv` remain but are orphaned
and inert (the chart no longer references them). The migration-skip is unit-covered by
`TestPluginsStorageMigration_Applies_BakedSignal`, so this case can be reasoned about rather than run.

### Minimum I actually recommend

**Tier 0 always**, plus **Tier 1 Case 1** on the UTM VM. Case 2/3 are the unchanged/heavy paths and are
unit-covered; skip them if resources are tight. CI runs the full Linux build + unit suite for
`internal/bll/blocknode` (which can't compile on macOS due to linux-only `internal/mount`).

## Risks

- The load-bearing behavior is the explicit `plugins.names: ""` (Helm re-applies the chart's non-empty
  default if the provisioner's `-f` output omits the key). Covered by
  `TestPluginsBakedTwoLayerMerge_ChartSeesEmpty`.
- **Upgrade path:** an existing managed install upgraded to a baked image leaves an orphaned (inert)
  plugins PV/PVC. Non-baked upgrades are unchanged (gate returns managed=true → identical to today).
- No CLI flag added or changed → no `docs/quickstart.md` change.

### DIFF
--- a/internal/bll/blocknode/install_handler.go
+++ b/internal/bll/blocknode/install_handler.go
@@ -56,7 +56,8 @@ func (h *InstallHandler) BuildWorkflow(
 	}
 
 	// Fail fast if storage paths can't be resolved — don't set up a cluster for nothing.
-	if err := bnpkg.ValidateStorageCompleteness(inputs.Custom.Storage, inputs.Custom.ChartVersion); err != nil {
+	if err := bnpkg.ValidateStorageCompleteness(inputs.Custom.Storage, inputs.Custom.ChartVersion,
+		!bnpkg.EffectivePluginsNamesEmpty(inputs.Custom.PluginList, inputs.Custom.ValuesFile)); err != nil {
 		return nil, err
 	}
 
--- a/internal/bll/blocknode/reconfigure_handler.go
+++ b/internal/bll/blocknode/reconfigure_handler.go
@@ -49,7 +49,8 @@ func (h *ReconfigureHandler) BuildWorkflow(
 	}
 
 	// Fail fast if storage paths can't be resolved.
-	if err := bnpkg.ValidateStorageCompleteness(inputs.Custom.Storage, inputs.Custom.ChartVersion); err != nil {
+	if err := bnpkg.ValidateStorageCompleteness(inputs.Custom.Storage, inputs.Custom.ChartVersion,
+		!bnpkg.EffectivePluginsNamesEmpty(inputs.Custom.PluginList, inputs.Custom.ValuesFile)); err != nil {
 		return nil, err
 	}
 
--- a/internal/bll/blocknode/upgrade_handler.go
+++ b/internal/bll/blocknode/upgrade_handler.go
@@ -84,7 +84,8 @@ func (h *UpgradeHandler) BuildWorkflow(
 	}
 
 	// Fail fast if storage paths can't be resolved.
-	if err := bnpkg.ValidateStorageCompleteness(inputs.Custom.Storage, inputs.Custom.ChartVersion); err != nil {
+	if err := bnpkg.ValidateStorageCompleteness(inputs.Custom.Storage, inputs.Custom.ChartVersion,
+		!bnpkg.EffectivePluginsNamesEmpty(inputs.Custom.PluginList, inputs.Custom.ValuesFile)); err != nil {
 		return nil, err
 	}
 
--- a/internal/blocknode/migration_storage.go
+++ b/internal/blocknode/migration_storage.go
@@ -33,6 +33,10 @@ const (
 	ctxKeyManager    = "blocknode.manager"
 	ctxKeyProfile    = "blocknode.profile"
 	ctxKeyValuesFile = "blocknode.valuesFile"
+	// ctxKeyManagePlugins carries the #913 plugins-baked signal (a bool) into
+	// StorageMigration.Applies. It is set before migration filtering, unlike
+	// ctxKeyManager which is only available for Execute/Rollback.
+	ctxKeyManagePlugins = "blocknode.managePlugins"
 )
 
 // StorageMigration handles the breaking change pattern where a new storage PV/PVC
@@ -75,7 +79,27 @@ func (m *StorageMigration) Applies(mctx *migration.Context) (bool, error) {
 	}
 
 	targetVersion, _ := mctx.Data.String(migration.CtxKeyTargetVersion)
-	return m.storage.RequiredByVersion(targetVersion), nil
+	if !m.storage.RequiredByVersion(targetVersion) {
+		return false, nil
+	}
+
+	// #913: the plugins storage is only managed when the effective plugins.names
+	// is non-empty. For a plugins-baked image (empty names) the chart renders no
+	// plugins volume, so an upgrade must not (re-)provision the plugins PV/PVC —
+	// otherwise the empty managed PVC would shadow the image's baked plugins.
+	// The signal is read from a context bool set by BuildMigrationWorkflow before
+	// migration filtering (the Manager itself is only added to the context later,
+	// for Execute/Rollback). When the key is absent (e.g. unit tests exercising
+	// pure version logic) the migration stays applicable — the pre-#913 default.
+	if m.storage.Name == "plugins" {
+		if v, ok := mctx.Data.Get(ctxKeyManagePlugins); ok {
+			if managePlugins, isBool := v.(bool); isBool && !managePlugins {
+				return false, nil
+			}
+		}
+	}
+
+	return true, nil
 }
 
 // Execute creates the storage directory on the host and the PV/PVC in Kubernetes.
--- a/internal/blocknode/migrations.go
+++ b/internal/blocknode/migrations.go
@@ -61,6 +61,10 @@ func BuildMigrationWorkflow(manager *Manager, profile, valuesFile string) (*auto
 	}
 	mctx.Data.Set(migration.CtxKeyInstalledVersion, installedVersion)
 	mctx.Data.Set(migration.CtxKeyTargetVersion, manager.blockNodeInputs.ChartVersion)
+	// #913: expose the plugins-baked signal to StorageMigration.Applies. It must be
+	// set before GetApplicableMigrations (which evaluates Applies); the Manager
+	// itself is only added to the context afterwards, for Execute/Rollback.
+	mctx.Data.Set(ctxKeyManagePlugins, manager.managesPluginsStorage())
 
 	migrations, err := migration.GetApplicableMigrations(migration.ScopeBlockNode, mctx)
 	if err != nil {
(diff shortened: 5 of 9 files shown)


brunodam (approved):
LGTM!

PR #938, opened 2026-07-29
feat(blocknode): persist last-chosen firewall & traffic-shaping values in state

### LINKED ISSUE #932: Persist last-chosen firewall & traffic-shaping values in state.yaml for faithful reconfigure/upgrade re-assert
## Summary

Persist the operator's **last-chosen deployment values** for the host firewall and the traffic-shaping bundle into `state.yaml`, and seed `block node reconfigure` / `upgrade` from them (below explicit flags/config). Today these values live only in the operator-supplied `--config` file and/or the live host (nft table, tc qdiscs), so they cannot be faithfully re-asserted when the operator re-runs `reconfigure`/`upgrade` without re-supplying them — most visibly on a **disable -> re-enable** cycle.

This is a design/enhancement follow-up split out of #929 (the two bug fixes there are unrelated and already handled). It deliberately does **not** try to store these in `config.yaml`: that file is operator-authored *input* for the tool, whereas these are machine-managed records of *what was chosen/deployed* — which is what `state.yaml` is for.

## Background / current behavior

- The host firewall config (`managementCidrs`, `blockedCidrs`, `sshPort`, `podCidr`, `inClusterPorts`) is read only from `config.Get().Host`, i.e. the `--config` file. `config.Initialize` loads **only** the `--config` path (default `""`); there is no auto-discovered/default config file on disk.
- `NetworkFirewallCreate` is create-if-missing (`Create(..., force=false)`): if the `inet host` table already exists it is a no-op and the **live nft table** is the de-facto store of the CIDRs. After a teardown, that store is gone.
- The tc-shaping *content* (`EgressInterface`, `LinkRate`, `ShapeOverrides`) is carried only on `BlockNodeInputs` for the current invocation; it is **not** persisted anywhere. `upgrade` re-asserts shaping with an empty NIC (auto-detect) and empty rate, rather than reproducing the operator's original values.
- Only the *decision* `TrafficShapingDisabled` is persisted (on `BlockNodeState`), because it is a pure decision with no cluster/config footprint. Its *content* siblings were never added.

### The gap (concrete)

```
# 1. install with an allowlist
block node install --firewall-enabled=true --mgmt-cidrs=10.0.0.0/8
# 2. disable -> tears the inet host table down
block node reconfigure --firewall-enabled=false
# 3. re-enable WITHOUT re-supplying CIDRs (and none in --config)
block node reconfigure --firewall-enabled=true
#    -> len(ManagementCIDRs)==0 guard fires -> firewall SKIPPED to avoid SSH lock-out.
#       The original allowlist is lost.
```

The same class of gap applies to tc content on `upgrade`/`reconfigure` re-assert.

## Why `state.yaml` (not `config.yaml`, not a second config file)

- `config.yaml` should remain **operator-authored input** (logging level, profile, desired deployment config). It is not where the tool should write back machine-managed "last chosen" values.
- `state.yaml` is already the durable record of **effective deployment values** and is a first-class resolution source. The RSL resolver (`internal/rsl/effective.go`) has an explicit `StrategyState` tier, and `BlockNodeState` already persists `ChartRef`, `Storage.BasePath`, `PluginPreset`, `PluginList`, retention, and `TrafficShapingDisabled`. Adding the firewall/tc content is the same established pattern.
- Per-field precedence is already field-specific in the resolver, so an explicit flag/`--config` value still wins; state only supplies the value when the operator didn't. This cleanly avoids config-vs-state drift.

## Proposed approach

1. Add the fields to `state.yaml`, respecting scope:
   - **Host firewall** -> `MachineState` (host-scoped; outlives any single block node).
   - **tc content** (`EgressInterface`, `LinkRate`, `ShapeOverrides`) -> `BlockNodeState` (part of the shaping bundle).
2. Persist the resolved values via the existing patch helpers on `install` / `reconfigure` (the same place `TrafficShapingDisabled` is written by `patchBlockNodeStateWithTrafficShaping`).
3. Wire them as a `StrategyState` source in the resolver, with per-field precedence so an explicit flag/config still wins; state is the fallback.
4. **Preserve them across reality refresh.** See caveats — this is the critical gotcha.
5. On re-enable, `NetworkFirewallCreate` (and the tc steps) resolve CIDRs/NIC/rate from flag > config > state, so a bare re-enable restores the last-known allowlist instead of skipping.

## State-schema sketch

> Illustrative only - names/polarity to be finalized in review.

```go
// internal/state/state.go

type MachineState struct {
    Profile  string                   `yaml:"profile,omitempty"`
    Software map[string]SoftwareState `yaml:"software"`
    Hardware map[string]HardwareState `yaml:"hardware"`
    // NEW: last-resolved host firewall config, so reconfigure/upgrade can
    // re-assert the inet host table (esp. after a disable->re-enable) without
    // the operator re-supplying --mgmt-cidrs. Host-scoped: lives here, not on
    // BlockNodeState, so it survives a block-node uninstall.
    Firewall *HostFirewallState `yaml:"firewall,omitempty" json:"firewall,omitempty"`
    LastSync htime.Time               `yaml:"lastSync,omitempty"`
}

// HostFirewallState mirrors models.HostConfig's operator-supplied fields.
// Populated from the resolved effective config on install/reconfigure; read
// back as the StrategyState fallback below flag/config.
type HostFirewallState struct {
    ManagementCIDRs []string `yaml:"managementCidrs,omitempty" json:"managementCidrs,omitempty"`
    BlockedCIDRs    []string `yaml:"blockedCidrs,omitempty"    json:"blockedCidrs,omitempty"`
    SSHPort         int      `yaml:"sshPort,omitempty"         json:"sshPort,omitempty"`
    PodCIDR         string   `yaml:"podCidr,omitempty"         json:"podCidr,omitempty"`
    InClusterPorts  []int    `yaml:"inClusterPorts,omitempty"  json:"inClusterPorts,omitempty"`
}

type BlockNodeState struct {
    // ... existing fields ...
    TrafficShapingDisabled bool `yaml:"trafficShapingDisabled,omitempty"`
    // NEW: last-resolved tc-shaping content, so upgrade/reconfigure re-assert
    // the same NIC/rate/overrides instead of falling back to auto-detect.
    Shaping *ShapingState `yaml:"shaping,omitempty" json:"shaping,omitempty"`
    LastSync htime.Time    `yaml:"lastSync,omitempty"`
}

type ShapingState struct {
    EgressInterface string                        `yaml:"egressInterface,omitempty" json:"egressInterface,omitempty"`
    LinkRate        string                        `yaml:"linkRate,omitempty"        json:"linkRate,omitempty"`
    ShapeOverrides  map[string]models.ShapeOverride `yaml:"shapeOverrides,omitempty" json:"shapeOverrides,omitempty"`
}
```

### Persist + resolve wiring (touch-points)

- `internal/bll/blocknode/helpers.go` - extend `patchBlockNodeStateWithTrafficShaping` (install + reconfigure) to write `BlockNodeState.Shaping`; add an equivalent machine-state patch (or profile-extractor sibling) for `MachineState.Firewall`.
- `internal/rsl/*` - register the new state fields as `StrategyState` sources with per-field precedence (flag > env > config > state > default), mirroring the existing storage/retention resolvers.
- `cmd/cli/commands/common/host_firewall.go` / `traffic_shaping.go` - seed the effective config from the resolved value (already partly done for the enable/disable *decision*; extend to the *content*).

## Caveats to decide up front

1. **Reality-refresh preservation (critical).** Any state field not observable from the cluster/Helm must be carried through the reality checkers' rebuilds, or a refresh silently resets it - the exact bug fixed for `TrafficShapingDisabled` in the block-node checker.
   - `internal/reality/blocknode_checker.go` **rebuilds `BlockNodeState` from scratch** when a Helm release is found -> `BlockNodeState.Shaping` MUST be explicitly preserved (copy from the prior `sm.State()` value), like `TrafficShapingDisabled`.
   - `internal/reality/machine_checker.go` starts from `current.MachineState` and only overwrites `Software`/`Hardware`/`LastSync` -> `MachineState.Firewall` is **auto-preserved**; still add a test to lock that in.
2. **Precedence:** confirm explicit flag/`--config` always beats persisted state (state is a fallback, not an override).
3. **Polarity/shape:** `TrafficShapingDisabled` uses negative polarity; new content fields should use positive/explicit values. Consider pointer/omitempty so "never set" is distinguishable from "set empty".
4. **Placement:** firewall is host-scoped -> `MachineState` (survives block-node uninstall); tc content is block-node-scoped -> `BlockNodeState`.
5. **Migration:** additive, `omitempty` fields -> old state files parse fine (absent = fall back to config/flag/auto-detect, i.e. today's behavior). No migration boundary needed.

## Acceptance criteria

- [ ] After `install --firewall-enabled=true --mgmt-cidrs=X`, `state.yaml` records the firewall config under `machineState.firewall`.
- [ ] `reconfigure --firewall-enabled=false` then `reconfigure --firewall-enabled=true` (no `--mgmt-cidrs`, no `--config`) restores the original allowlist instead of skipping with the SSH-lockout guard.
- [ ] `upgrade` on a shaping-enabled node re-asserts the **same** `EgressInterface`/`LinkRate`/overrides recorded at install, not auto-detected values.
- [ ] An explicit flag or `--config` value still overrides the persisted state value.
- [ ] A reality refresh (handler `Refresh` + `FlushState`) does not drop the new fields (regression tests for both checkers).
- [ ] Old `state.yaml` files (without the new fields) load and behave exactly as today.

## References

- PR #929 (feat: enable/disable firewall & traffic shaping on reconfigure/upgrade) - Risks section already notes "teardown rollbacks are no-ops: re-enabling a feature needs inputs the teardown steps don't snapshot." This issue closes that gap.
- Related fixes on the #929 branch: reality-checker preservation of `TrafficShapingDisabled`, and egress-NIC unshape on teardown.

alex-au:
## Summary

Persist the operator's **last-chosen host-firewall config** and **traffic-shaping content** into `state.yaml`, and seed `block node reconfigure` / `upgrade` from them (below explicit flags/config). Today the firewall CIDRs live only in the `--config` file or the live nft table, and the tc content (`EgressInterface`/`LinkRate`/`ShapeOverrides`) is carried per-invocation and never persisted — so a `disable → re-enable` cycle loses the allowlist (the SSH-lockout guard skips the firewall) and `upgrade` re-asserts shaping with an auto-detected NIC / empty rate instead of the original values.

Design follow-up split out of #929 (no epic). Values go into `state.yaml` (machine-managed record of what was deployed), **not** `config.yaml` (operator-authored input).

### What changed

- **State schema** (`internal/state/state.go`): add `MachineState.Firewall *HostFirewallState` (host-scoped — survives a block-node uninstall) and `BlockNodeState.Shaping *ShapingState`. Both are pointer + `omitempty`, so old `state.yaml` files load unchanged.
- **Firewall enable/disable decision**: `HostFirewallState.Disabled` (negative polarity, mirroring `TrafficShapingDisabled` / `HostConfig.Disabled`). The install/reconfigure patch always records the decision but updates the CIDR/port content **only when enabled with a non-empty allowlist**, so a disable — or an enable with an empty allowlist, which the step skips — flips `Disabled` while preserving the last-known-good allowlist for a future bare re-enable.
- **Traffic-shaping content**: persisted on install/reconfigure when enabled; the shared effective-inputs resolver falls back to `BlockNodeState.Shaping` when the operator supplies no egress NIC / rate / overrides, so `upgrade` (which has no such flags) and a bare `reconfigure` re-assert the original shaping. Explicit input always wins.
- **Resolution precedence**: host-firewall CIDRs resolve `flag > config > state > default`; `upgrade` seeds the firewall from state gated on the persisted `Disabled` decision (re-assert when enabled, skip when disabled).
- **Reality-refresh preservation**: `BlockNodeState.Shaping` is copied through the block-node checker's found-release rebuild (exactly like `TrafficShapingDisabled`); `MachineState.Firewall` is preserved by the machine checker (it starts from the current `MachineState`).

No new CLI flags or subcommands, so `docs/quickstart.md` is unchanged.

### Acceptance criteria

- [x] After `install --firewall-enabled=true --mgmt-cidrs=X`, `state.yaml` records the firewall under `machineState.firewall`.
- [x] `reconfigure --firewall-enabled=false` then `reconfigure --firewall-enabled=true` (no `--mgmt-cidrs`, no `--config`) restores the original allowlist instead of skipping with the SSH-lockout guard.
- [x] `upgrade` re-asserts the same `EgressInterface`/`LinkRate`/overrides recorded at install, not auto-detected values.
- [x] An explicit flag or `--config` value still overrides the persisted state value.
- [x] A reality refresh does not drop the new fields (regression tests for both checkers).
- [x] Old `state.yaml` files (without the new fields) load and behave exactly as today.

> **Note on scope beyond the issue:** the issue's schema sketch and ACs cover firewall re-assert on `reconfigure` and tc re-assert on `upgrade`. This PR additionally persists a firewall enable/disable **decision** so `upgrade` can *also* faithfully re-assert the firewall — plain create-if-missing with no persisted decision would either no-op or wrongly re-create a firewall the operator had disabled.

## Test plan

Unit (`internal/state` and `internal/reality` run on macOS; `internal/bll/blocknode` and the `cmd` packages require the UTM VM due to the Linux-only `mount` dependency):

```bash
task vm:test:unit
# or, targeted:
go test -tags='!integration' ./internal/state/... ./internal/reality/... ./internal/bll/blocknode/... ./cmd/cli/commands/common/...
```

New/updated tests:
- `internal/state`: schema round-trip + old-file load (`state_firewall_shaping_test.go`), firewall doc parsing (`state_reader_test.go`).
- `internal/reality`: `TestRefreshState_PreservesShaping`, `TestMachineRefreshState_PreservesFirewall`.
- `internal/bll/blocknode`: patch writes firewall/shaping, skip-on-disable and skip-on-empty preserve the allowlist (`helpers_patch_test.go`); tc content falls back to state and explicit input wins (`effective_inputs_test.go`).

## Risks

- **Clobbering the allowlist on disable** — mitigated by the skip-on-disable / skip-on-empty persistence gate; covered by unit tests.
- **Silent reset on reality refresh** — the block-node checker rebuild is the known footgun (same class fixed for `TrafficShapingDisabled`); explicit copy + regression test.
- Additive `omitempty` pointer fields → old `state.yaml` loads unchanged; rollback is a pure revert (persisted fields ignored by older binaries).

Closes #932

### DIFF
--- a/cmd/cli/commands/block/node/reconfigure.go
+++ b/cmd/cli/commands/block/node/reconfigure.go
@@ -73,6 +73,16 @@ var (
 			// on --profile=local preempt the clearer "block node is not installed"
 			// precondition error on a fresh host.
 			if trafficShapingEnabled {
+				// Seed the egress prompts from the persisted last-chosen shaping
+				// content so a default-accept keeps the operator's previous NIC/link
+				// rate instead of reverting to auto-detection. An explicit CLI flag
+				// still wins (ResolveEgressConfig skips a prompt whose flag was set).
+				if !cmd.Flags().Changed(common.FlagNameEgressInterface) && flagEgressInterface == "" {
+					flagEgressInterface = stateDefaults.BlockNode.EgressInterface
+				}
+				if !cmd.Flags().Changed(common.FlagNameLinkRate) && flagLinkRate == "" {
+					flagLinkRate = stateDefaults.BlockNode.LinkRate
+				}
 				if err := common.ResolveEgressConfig(cmd, args, cv, &flagEgressInterface, &flagLinkRate); err != nil {
 					return err
 				}
--- a/cmd/cli/commands/block/node/upgrade.go
+++ b/cmd/cli/commands/block/node/upgrade.go
@@ -42,6 +42,17 @@ var (
 				return err
 			}
 
+			// Upgrade exposes no firewall flags and never prompts; seed the effective
+			// host-firewall config from the last-persisted state so the network-plane's
+			// NetworkFirewallCreate step re-asserts the operator's last-known allowlist
+			// (create-if-missing) instead of skipping on an empty one. The persisted
+			// enable/disable decision is honored — a firewall that was disabled stays
+			// off (the step self-gates on config.Host.Disabled). Traffic-shaping content
+			// is re-asserted separately via the shared effective-inputs resolver.
+			if err := common.SeedHostFirewallFromState(); err != nil {
+				return err
+			}
+
 			intent := models.Intent{
 				Action: models.ActionUpgrade,
 				Target: models.TargetBlockNode,
--- a/cmd/cli/commands/common/egress.go
+++ b/cmd/cli/commands/common/egress.go
@@ -55,6 +55,14 @@ func ValidateEgressFlags(cmd *cobra.Command, linkRate string) error {
 // interactive and the flags were not supplied on the CLI. egressInterface and
 // linkRate are updated in-place by the prompts.
 //
+// egressInterface and linkRate double as seed inputs: when either already holds
+// a non-empty value (e.g. `block node reconfigure` pre-seeding it from the
+// persisted BlockNodeState.Shaping last-chosen values), that value becomes the
+// prompt's pre-filled default so a default-accept keeps the operator's
+// previously-configured NIC / link rate instead of silently reverting to
+// auto-detection. On a fresh `install` both are empty, so the detected NIC and
+// sysfs link speed are used — the original behaviour.
+//
 // When cv is non-nil the prompted values are recorded into it and no separate
 // summary is printed — the caller is responsible for printing the unified
 // summary after all prompt sections complete. When cv is nil a local collector
@@ -76,8 +84,13 @@ func ResolveEgressConfig(
 		return nil
 	}
 
-	// Detect the egress NIC to use as the prompt default.
+	// Detect the egress NIC to use as the prompt default, unless a value was
+	// already seeded (persisted last-chosen NIC on reconfigure), which wins.
 	detectedNIC, _ := shape.DetectEgressInterface()
+	egressDefault := detectedNIC
+	if *egressInterface != "" {
+		egressDefault = *egressInterface
+	}
 
 	// Speed hint from sysfs: prefer the already-set flag value over the
 	// auto-detected one so the hint reflects the NIC the operator chose.
@@ -92,13 +105,23 @@ func ResolveEgressConfig(
 		}
 	}
 
+	// Seed the link-rate prompt from the persisted last-chosen rate (when the
+	// caller pre-seeded linkRate, e.g. reconfigure reading BlockNodeState.Shaping)
+	// so a default-accept keeps the operator's configured rate instead of
+	// reverting to the NIC's raw sysfs line speed. Falls back to the sysfs hint
+	// on a fresh install where nothing was chosen yet.
+	linkRateDefault := speedHint
+	if *linkRate != "" {
+		linkRateDefault = *linkRate
+	}
+
 	localCV := cv
 	if localCV == nil {
 		localCV = prompt.NewChosenValues()
 	}
 	if err := prompt.RunInputPrompts(cmd, []prompt.InputPrompt{
-		prompt.EgressInterfaceInputPrompt(detectedNIC, speedHint, egressInterface),
-		prompt.LinkRateInputPrompt(speedHint, linkRate),
+		prompt.EgressInterfaceInputPrompt(egressDefault, speedHint, egressInterface),
+		prompt.LinkRateInputPrompt(linkRateDefault, linkRate),
 	}, localCV); err != nil {
 		return err
 	}
--- a/cmd/cli/commands/common/host_firewall.go
+++ b/cmd/cli/commands/common/host_firewall.go
@@ -6,7 +6,9 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/automa-saga/logx"
 	"github.com/hashgraph/solo-weaver/internal/network/firewall"
+	"github.com/hashgraph/solo-weaver/internal/state"
 	"github.com/hashgraph/solo-weaver/internal/ui/prompt"
 	"github.com/hashgraph/solo-weaver/pkg/config"
 	"github.com/hashgraph/solo-weaver/pkg/models"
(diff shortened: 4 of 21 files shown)


brunodam (dismissed):
LGTM

brunodam (approved):

PR #941, opened 2026-07-29
test(daemon): assert NetworkUpgradeExecute GVR group matches installe…

### LINKED ISSUE #851: Add integration test asserting GVR group matches installed CRD
**Epic:** #1076

Add a test (or test helper) that queries the cluster for the `networkupgradeexecutes` CRD and asserts that the `Group` field in the hardcoded GVR constant matches the CRD's actual `.spec.group`. This prevents the API group from drifting again.

## Acceptance Criteria
- [ ] Test fails if GVR group does not match CRD spec group
- [ ] Test is gated on a real cluster being available (integration tag or `//go:build integration`)

radkomih:
## Description

* Add a `require_cluster` integration test (`TestWithCluster_UpgradeMonitor_GVRMatchesInstalledCRD`) that asserts the monitor's `NetworkUpgradeExecute` GVR (group, resource, served version) matches the CRD solo-operator actually installs. The CRD is looked up by plural name, so a wrong group still resolves the installed CRD and the assertion fails — catching real API-group drift rather than silently skipping.
* Export `consensus.NetworkUpgradeExecuteGVR()` so the test asserts against the production GVR value instead of a copy.
* Enable `SoloOperator` in `Test_ClusterSetup` so its CRDs are provisioned in Phase 2, making the guard a real assertion (not a skip) in Phase 3.
* Refresh the solo-operator CRD lists in the cluster health check (`step_health.go`) and UAT task (`uat.yaml`) to the 0.3.2 set (drop `networkoperations`, add the `networkgeneses`/`networkupgrade*` CRDs).
* Bump solo-operator `0.3.1` → `0.3.2` in the infrastructure catalog (with checksum) and update the config integration-test expectation.

### Related Issues

* Closes #851

### DIFF
--- a/internal/daemon/consensus/upgrade_monitor.go
+++ b/internal/daemon/consensus/upgrade_monitor.go
@@ -33,6 +33,12 @@ var networkUpgradeExecuteGVR = schema.GroupVersionResource{
 	Resource: "networkupgradeexecutes",
 }
 
+// NetworkUpgradeExecuteGVR returns the GVR the monitor watches. Exported so
+// tests can assert it against the CRD solo-operator installs.
+func NetworkUpgradeExecuteGVR() schema.GroupVersionResource {
+	return networkUpgradeExecuteGVR
+}
+
 const (
 	// NetworkUpgradeExecute status.phase values are defined as the shared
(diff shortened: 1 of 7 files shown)


radkomih:
closing, the fix has already been merged

PR #955, opened 2026-07-31
refactor(network/shape): collapse egress render/apply pipeline and single-source tc encoding

### LINKED ISSUE #946: refactor(network/shape): collapse the six-method egress render/apply pipeline and single-source tc command encoding
## Problem

The egress render/apply pipeline in `internal/network/shape/manager.go` is spread across **six overlapping methods with near-synonymous names**, and it is the hardest control flow to follow in the shape package. You cannot tell from a call site which one to use:

- `renderAndApplyEgress` (detect NIC → renderAndApplyScript)
- `renderAndApplyUnshapeEgress` (teardown-only: direct `tc qdisc del` + restart)
- `renderAndApplyScript` (render → write → restart service)
- `renderEgressScript` (three-way branch: no device → default; device but no classes → derive-or-fallback; else from config)
- `writeEgressScript` (sha256-gated write)
- `persistEgressScript` (render → write, **no** restart — used by `set`)

The three-way branch inside `renderEgressScript` (`manager.go:601-625`) plus the "restart vs no-restart vs direct-qdisc-del" fan-out is effectively a small state machine expressed implicitly across these methods.

Separately, **tc commands are hand-encoded twice**: once as Go `[]string` args in `internal/network/shape/tc_linux.go` (`ClassAdd`, `ClassAddRoot`, `QdiscAddRoot`, …) and once as shell text in the boot-script template `internal/templates/files/network/solo-provisioner-tc-egress.sh.tmpl`. The two encodings of `tc class add … htb rate … ceil … prio …` can silently drift — a change to HTB arg ordering or a new field must be edited in both places with no compiler or test tying them together.

This is tech-debt from the traffic-shaper feature landing incrementally across ~10 PRs (#746, #895, #921, etc.); no single behavior is wrong, but the surface is harder to maintain and extend than it needs to be.

## Proposed fix

1. **Collapse the render/apply pipeline** into one `renderEgressScript` (pure: registry → script string) plus one `applyEgressScript(ctx, mode)` where `mode` is an explicit enum/flag over the three real outcomes:
   - `restart` — structural change (device/class create/delete): write script + restart oneshot
   - `persistOnly` — live `tc class change` already applied (`set`): write script, no restart
   - `unshape` — teardown: write empty script + direct `tc qdisc del root`

   This makes the "why no restart here?" reasoning that currently lives in scattered doc comments a single, legible decision.

2. **Single-source the tc command encoding.** Either (a) render the boot script by calling the same `TCRunner`-arg builders the live path uses (so there is one encoding), or (b) generate the template's `tc` lines from the same arg slices. Prefer (a) if feasible.

Behavior must be preserved exactly — the shape package has ~2,200 lines of tests (`shape_test.go`, `override_test.go`, `veth_test.go`) and golden-asserts rendered script output, so this is a mechanical, test-guarded refactor.

## Notes / dependencies

- **Do NOT bundle the legacy "SPEED" sysfs-fallback removal into this.** That path (`defaultScriptData` + the `SpeedMbit >= 0` template branch + three validator "legacy shell expression: skip" escape hatches) is still reachable via `network shape delete --device egress` (which deliberately re-applies stock default shaping) and via legacy on-disk registries with an unresolved `"auto"` rate. Retiring it first requires redefining `DeleteDevice(egress)` behavior (route it through the existing unshape render, matching `TeardownEgress`) and is a separate decision.
- Related sibling cleanup already landed on branch `refactor-traffic-shaper-engine` (egress/ingress default-profile dedup, dead-code removal, `TCRunner` interface de-duplication, validation dedup).

## Acceptance

- [ ] The egress render/apply pipeline is expressed as one pure render + one apply with an explicit mode, replacing the six-method tangle.
- [ ] tc command argument encoding exists in exactly one place (live path and boot script share it) — or a test asserts the two encodings stay in lockstep.
- [ ] All existing `internal/network/shape` unit tests pass unchanged (behavior-preserving); golden script output is identical or the golden files are updated with a clear rationale.
- [ ] No change to the legacy SPEED path in this PR (tracked separately).

alex-au:
## Description

Behavior-preserving cleanup of the egress render/apply pipeline in `internal/network/shape`, the follow-up refactor filed from #949. Two things get untangled:

1. **The six-method egress pipeline** in `manager.go` was spread across near-synonymous methods (`renderAndApplyEgress`, `renderAndApplyUnshapeEgress`, `renderAndApplyScript`, `persistEgressScript`, plus the pure `renderEgressScript` and the `writeEgressScript` primitive). You could not tell from a call site which to use, and the "restart vs no-restart vs direct-qdisc-del" fan-out was an implicit state machine. This collapses the four *pipeline* methods into **one `applyEgressScript(ctx, nic, mode)`** with an explicit `egressApplyMode` enum (`applyRestart` / `applyPersistOnly` / `applyUnshape`), keeping `renderEgressScript` (pure) and `writeEgressScript` (write primitive). The "why no restart here / why a direct qdisc-del" reasoning now lives in one place — the enum doc comments.

2. **The tc command encoding was hand-written twice** — as Go `[]string` args in `tc_linux.go` and as shell text in the boot-script template — with nothing tying the HTB argument order and field set together. This extracts the encoding into build-tag-free `tc*Args` builders in `tc.go` that the live path (`execTCRunner`) now executes, and adds a **lockstep test** asserting the boot script renders the same commands token-for-token, so the two encodings cannot silently drift.

No functional change. NIC resolution is preserved exactly: empty `nic` auto-detects (operator create/delete verbs, keeping the old detect-and-decorate error); an explicit `nic` is used verbatim (the provision path passing the operator-chosen `--egress-interface`, and `set` reusing the NIC it already detected for the live `tc class change` — no double-detect). The boot-script template and the deliberately out-of-scope legacy SPEED path are untouched; golden script output is identical.

### Stacking note

Stacked on `refactor-traffic-shaper-engine` (PR #949), **not** `main` — this is the structural follow-up #949's description points at. Review/merge after #949.

### Files changed

| File | Change |
|---|---|
| `internal/network/shape/manager.go` | Add `egressApplyMode` enum + single `applyEgressScript`; delete `renderAndApplyEgress` / `renderAndApplyUnshapeEgress` / `renderAndApplyScript` / `persistEgressScript`; update 8 internal call sites (messages unchanged). |
| `internal/network/shape/tc.go` | Add build-tag-free `tc*Args` arg-builders — the single source of the tc encoding. |
| `internal/network/shape/tc_linux.go` | `execTCRunner` methods call the shared builders; drop now-unused `strconv`. |
| `internal/network/shape/tc_encoding_test.go` (new) | Lockstep test: renders the boot script, asserts each `tc` line equals the arg-builder encoding (whitespace/quote-insensitive). |
| `internal/network/shape/shape_test.go` | Update 3 call sites + comments to the new API; rename 2 tests. |

### Review guide

Checklist:
- `manager.go` `applyEgressScript` — `applyPersistOnly` returns before `applyEgress` (no service restart); `applyUnshape` renders the *unshape* script → write → direct `QdiscDelRoot` → `applyEgress` (same order as the old `renderAndApplyUnshapeEgress`); `applyRestart` = write → `applyEgress`.
- NIC source: `SetClass` passes its already-detected `nic`; `ProvisionDefaultEgress` / `RenderAndApplyDefaultEgress` pass the caller's explicit NIC (must **not** auto-detect — matters on multi-NIC hosts); create/delete verbs pass `""`.
- `tc.go` builders reproduce the exact prior arg slices (`1:`+minor, `handle+":"`, `strconv.Itoa(prio)`); `solo-provisioner-tc-egress.sh.tmpl` is untouched.
- Exported surface unchanged (`RenderAndApplyDefaultEgress`, `ProvisionDefaultEgressShape`, `ProvisionDefaultIngressShape`, `TeardownEgress`).

Tests:
```bash
go test ./internal/network/shape/...
GOOS=linux go vet ./internal/network/shape/... && GOOS=linux go test -c -o /dev/null ./internal/network/shape/
task lint
```
All pass (84 on darwin incl. `TestBootScriptTcEncodingMatchesArgBuilders`); lint 0 issues; `go vet` clean.

Manual UAT (UTM VM):
1. **Install renders identically** — `block node install …`, then `cat /opt/solo/tc-egress.sh`. The rendered HTB hierarchy is byte-identical to a pre-change install (same `tc qdisc/class` lines, same rates).
2. **`set` does not restart the oneshot** — `network shape set --class partner --rate 200mbit`, then `systemctl show -p ExecMainStartTimestamp solo-provisioner-tc-egress`: the timestamp is unchanged (live `tc class change` applied, no qdisc churn), and `/opt/solo/tc-egress.sh` shows the new rate for reboot persistence.
3. **create/delete restarts and matches live** — `network shape create --device egress --rate 1gbit --default reserve-egress` then `network shape delete --device egress`: the oneshot restarts each time and `tc -s qdisc show dev <nic>` / `tc class show dev <nic>` match the registry.
4. **Teardown unshapes (not re-shapes)** — disable traffic shaping (BN uninstall / `TeardownEgress`), then `tc qdisc show dev <nic>`: the root qdisc is **dropped**, not replaced with the stock default HTB hierarchy.
5. **Encoding lockstep holds** — edit the HTB arg order in either `tc.go` or the template and run `go test ./internal/network/shape/...`: `TestBootScriptTcEncodingMatchesArgBuilders` fails, proving the two encodings are tied.

### Related Issues

* Closes #946

### DIFF
--- a/internal/network/shape/manager.go
+++ b/internal/network/shape/manager.go
@@ -122,7 +122,7 @@ func (m *Manager) CreateDevice(ctx context.Context, dev *DeviceConfig, force boo
 			return err
 		}
 		if dev.Dir == DirEgress {
-			if err := m.renderAndApplyEgress(ctx); err != nil {
+			if err := m.applyEgressScript(ctx, "", applyRestart); err != nil {
 				return err
 			}
 		}
@@ -186,7 +186,7 @@ func (m *Manager) CreateClass(ctx context.Context, cls *ClassConfig, force bool)
 			return err
 		}
 		if ci.Dir == DirEgress {
-			if err := m.renderAndApplyEgress(ctx); err != nil {
+			if err := m.applyEgressScript(ctx, "", applyRestart); err != nil {
 				return err
 			}
 		}
@@ -258,8 +258,9 @@ func (m *Manager) SetClass(ctx context.Context, name string, rate, ceil *string,
 			}
 			// Persist the boot script for reboot without restarting the service:
 			// the live tc class change above already updated the kernel, and a
-			// restart would tear down and rebuild the root qdisc.
-			if err := m.persistEgressScript(nic); err != nil {
+			// restart would tear down and rebuild the root qdisc. Reuse the NIC
+			// detected above rather than re-detecting.
+			if err := m.applyEgressScript(ctx, nic, applyPersistOnly); err != nil {
 				return errorx.Decorate(err,
 					"live tc class change applied but boot script re-render failed; reboot may revert the change")
 			}
@@ -405,7 +406,7 @@ func (m *Manager) DeleteClass(ctx context.Context, name string) error {
 			return err
 		}
 		if ci.Dir == DirEgress {
-			if err := m.renderAndApplyEgress(ctx); err != nil {
+			if err := m.applyEgressScript(ctx, "", applyRestart); err != nil {
 				return errorx.Decorate(err,
 					"class config deleted but boot script re-render failed; restart tc-egress.service to sync")
 			}
@@ -446,7 +447,7 @@ func (m *Manager) DeleteDevice(ctx context.Context, dir string) error {
 		}
 		if dir == DirEgress {
 			// Re-render with default (empty) egress config.
-			if err := m.renderAndApplyEgress(ctx); err != nil {
+			if err := m.applyEgressScript(ctx, "", applyRestart); err != nil {
 				return errorx.Decorate(err,
 					"device config deleted but boot script re-render failed; restart tc-egress.service to sync")
 			}
@@ -482,11 +483,11 @@ func (m *Manager) TeardownEgress(ctx context.Context) error {
 		}
 		// Re-render an unshape boot script (delete root qdisc, re-add nothing) and
 		// apply it so both the boot script and the live kernel hierarchy are reset
-		// to unshaped. This must NOT go through renderAndApplyEgress: with the
-		// device/class registry now empty, that path falls through to the default
-		// shaping render and would re-apply the stock HTB hierarchy instead of
-		// removing it.
-		if err := m.renderAndApplyUnshapeEgress(ctx); err != nil {
+		// to unshaped. applyUnshape (not applyRestart) is required: with the
+		// device/class registry now empty, the normal render falls through to the
+		// default shaping render and would re-apply the stock HTB hierarchy instead
+		// of removing it.
+		if err := m.applyEgressScript(ctx, "", applyUnshape); err != nil {
 			return errorx.Decorate(err,
 				"egress shape config removed but boot script re-render failed; restart tc-egress.service to sync")
 		}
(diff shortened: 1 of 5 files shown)


brunodam (approved):
LGTM

PR #961, opened 2026-08-04
fix(daemon/blocknode): statusz poll loop is silent — add per-tick logging

### LINKED ISSUE #958: fix(daemon/blocknode): statusz poll loop is silent — add per-tick logging
## Problem

The traffic-shaper monitor's statusz poll loop (`runStatuszPoll` in `internal/daemon/blocknode/traffic_shaper_monitor.go`) emits **no log entry** on steady-state ticks. The only log lines that fire are on URL transitions (endpoint resolved / endpoint lost). Once the loop settles at a configured URL, every subsequent tick — whether it successfully checks statusz, skips the apply because the digest is unchanged, or fires a privileged apply — is completely silent.

This makes it impossible to tell from daemon logs whether the daemon is polling at all. Combined with the separate block-node-side issue (the BN statusz endpoint not loading new config), there is no observability into whether the daemon–BN polling cycle is functioning.

## Affected code

`internal/daemon/blocknode/traffic_shaper_monitor.go`, `reconcile` closure in `runStatuszPoll` (~line 274):

```go
// No log here — every tick is silent once the URL is set
digest, err := m.delegator.ReconcileShaperCheck(ctx, statuszURL)
if err != nil {
    return err
}
if digest == lastDigest && !lastApply.IsZero() && time.Since(lastApply) < statuszForceResyncInterval {
    return nil  // silent skip
}
if err := m.delegator.ReconcileShaper(ctx, statuszURL); err != nil {
    return err  // silent apply attempt
}
```

## Proposed fix

Add three log entries to `reconcile()`:

| Point | Level | `reason` field | When |
|---|---|---|---|
| Before `ReconcileShaperCheck` | `debug` | `TrafficShaperStatuszPolling` | Every tick with a configured URL |
| Digest-gate skip branch | `debug` | `TrafficShaperStatuszUnchanged` | Digest unchanged, apply skipped |
| Before `ReconcileShaper` | `info` | `TrafficShaperStatuszApplying` | Membership changed or force-resync |

The `debug` entries are suppressed at the default log level and visible with `--log-level debug`; the `info` apply entry always fires so state changes appear in normal operation.

## Impact

- Without the per-tick log, operators cannot verify the daemon is polling statusz during troubleshooting.
- The apply log is especially important: without it there is no daemon-side record of when nft policy membership was last reconciled.

alex-au:
## Description

The traffic-shaper monitor's statusz poll loop (`runStatuszPoll` → `reconcile` in `internal/daemon/blocknode/traffic_shaper_monitor.go`) only logged on endpoint transitions (resolved/lost). Once the loop settled on a URL, every subsequent tick — a successful check, a digest-gated skip, or a privileged apply — was completely silent, so operators had no way to tell from daemon logs whether the daemon–BN polling cycle was running at all.

This adds three log entries inside `reconcile()`, exactly as specified in the issue:

| Point | Level | `reason` | Fires |
|---|---|---|---|
| Before `ReconcileShaperCheck` | `debug` | `TrafficShaperStatuszPolling` | Every tick with a configured URL |
| Digest-gate skip branch | `debug` | `TrafficShaperStatuszUnchanged` | Digest unchanged, apply skipped |
| Before `ReconcileShaper` | `info` | `TrafficShaperStatuszApplying` | Membership changed or force-resync |

The two `debug` lines are suppressed at the default log level; the `info` apply line always fires so nft policy reconciliation is visible in normal operation. All three carry `monitor` and `statusz_url` fields, matching the existing `TrafficShaperStatuszEndpointResolved`/`...Lost` lines in the same function. This is a pure logging change — no control-flow or behavior change.

### Files changed

| File | Change |
|---|---|
| `internal/daemon/blocknode/traffic_shaper_monitor.go` | Add the three log entries in `reconcile()` inside `runStatuszPoll` |

### Review guide

**Invariants to verify**
- `TrafficShaperStatuszPolling` fires on every tick with a non-empty `statuszURL`, before the check exec — including the entry reconcile.
- `TrafficShaperStatuszUnchanged` fires only inside the existing skip-gate branch (digest unchanged AND `lastApply` non-zero AND within `statuszForceResyncInterval`) — never on the entry apply or a digest change.
- `TrafficShaperStatuszApplying` fires immediately before `ReconcileShaper` — i.e. on the entry apply, a digest change, and a forced resync — matching every place the existing tests increment `applyCalls`.
- No new log line fires when `statuszURL == ""` (idle loop stays quiet, unchanged from before).
- No behavioral change to the digest-gate or apply logic — logging only.

**Test commands**
```bash
task test:unit TEST_PATHS=./internal/daemon/blocknode/...
# expect: ok  github.com/hashgraph/solo-weaver/internal/daemon/blocknode

task lint
# expect: 0 issues.
```

**Manual UAT**
1. Build the daemon and install it against a cluster with a block node running, with `--log-level debug` (or `log_level: debug` in the daemon config).
2. Tail the daemon log and confirm, once per poll interval (default 5s):
   ```
   ... reason=TrafficShaperStatuszPolling ... msg="polling statusz"
   ```
   followed by either
   ```
   ... reason=TrafficShaperStatuszUnchanged ... msg="desired membership digest unchanged — skipping apply"
   ```
   or
   ```
   ... reason=TrafficShaperStatuszApplying ... msg="applying nft policy membership from statusz"
   ```
3. Restart the daemon at the default (info) log level and confirm the two `debug` lines (`TrafficShaperStatuszPolling`/`...Unchanged`) no longer appear, but `TrafficShaperStatuszApplying` still fires whenever BN roster/policy membership changes (e.g. scale the BN deployment or edit its plugin config to force a digest change).
4. Confirm the pre-existing `TrafficShaperStatuszEndpointResolved`/`...Lost` transition logs are unaffected (e.g. delete/recreate the watched BN pod).

**Risks / rollback**
- Additive logging only; no control-flow, template, or upgrade-path changes.

### Related Issues

* Closes #958


brunodam (approved):
LGTM

PR #972, opened 2026-08-05
ci: run UAT and integration jobs in parallel

brunodam:
## Description

Two independent improvements bundled here to avoid adding CI load.

### CI parallelism

The UAT and integration-test jobs in `flow-pull-request-checks.yaml` were gated in a fixed sequence: each UAT job waited for `uat-compat` to finish, and the integration-test job waited for all three UAT jobs. Since these suites are independent of each other (each still requires the `build` job), the sequencing only added wait time without providing any ordering guarantee that matters.

This change removes the redundant `needs:` dependencies so `uat-core`, `uat-purge-storage`, `uat-plugin-preset-upgrade`, and the integration-test job all start as soon as `build` is green, cutting CI wall time proportionally to the slowest suite.

### Helm timeout diagnostics

After a `--atomic` Helm install/upgrade timeout, `blockNodeChartError` now makes a live Kubernetes API call to fetch pod container states and recent Warning events, appending them to the resolution hints shown to the operator. Instead of static `kubectl` command suggestions, the operator sees the actual failure reason (e.g. `ImagePullBackOff`, `CrashLoopBackOff`) and the relevant event messages directly in the error output.

`KindEvent` is registered in the kube client's GVR map so the dynamic client can list events by field selector. `PodDiagnostics` suppresses all internal errors so a cluster-unreachable diagnostic never masks the original Helm timeout error.

### Related Issues

None.

### DIFF
--- a/.github/workflows/flow-pull-request-checks.yaml
+++ b/.github/workflows/flow-pull-request-checks.yaml
@@ -79,7 +79,6 @@ jobs:
     needs:
       - changes
       - build
-      - uat-compat
     if: needs.changes.outputs.non_docs == 'true'
     with:
       custom-job-label: UAT Core
@@ -92,7 +91,6 @@ jobs:
     needs:
       - changes
       - build
-      - uat-compat
     if: needs.changes.outputs.non_docs == 'true'
     with:
       custom-job-label: UAT Purge Storage
@@ -105,7 +103,6 @@ jobs:
     needs:
       - changes
       - build
-      - uat-compat
     if: needs.changes.outputs.non_docs == 'true'
     with:
       custom-job-label: UAT Plugin Preset Upgrade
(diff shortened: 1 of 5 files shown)


.github/workflows/flow-pull-request-checks.yaml
```diff
@@ -72,55 +72,53 @@ jobs:
...
       - build
-      - uat-compat
     if: needs.changes.outputs.non_docs == 'true'
     with:
       custom-job-label: UAT Plugin Preset Upgrade
       uat-scenario: plugin-preset-upgrade
       vm-os: debian
+      vm-cpus: 2
 
   integration-tests:
```
alex-au, line 116: this is the slowest CI among all, although we removed the dependencies from this test, but it will still be the last test to start if all runners are occupied, should we place it right after the unit test to ensure it starts early so less overall time on CIs?

alex-au (approved):
LGTM, just a nit, feel free to ignore.

.github/workflows/flow-pull-request-checks.yaml
brunodam: I'll add that in the next PR

PR #974, opened 2026-08-05
feat(cli): resolve the daemon binary automatically and complete the uninstall teardown

### LINKED ISSUE #973: fix(cli): resolve the daemon binary automatically instead of prompting for --daemon-bin on --profile=local
## Problem

Installing a block node with traffic shaping enabled under `--profile=local` always
stops to ask for the daemon binary path in the TUI:

```
Daemon Binary Path (required for --profile=local)
Path to a locally-built solo-provisioner-daemon binary. Local/dev builds have no
downloadable release to auto-download.
> bin/solo-provisioner-daemon-linux-amd64
```

`resolveDaemonBinarySource` (`cmd/cli/commands/block/node/daemon_offer.go:66`) prompts
whenever `profile == local` and `--daemon-bin` is empty. Three things are wrong with
that gate:

**1. It prompts even when nothing will be installed.** `install.go:70` resolves the
source during the prompt phase, but its only consumer — `ensureBlockNodeDaemon`
(`daemon_offer.go:43`) — returns early when the daemon service is already running. In
the normal dev loop (daemon already up, re-installing the block node) the operator
answers a question whose answer is then discarded.

**2. It prompts even when the daemon binary is already on the host.** A previous
`daemon service install` (or an earlier block-node install) leaves a working binary at
`/opt/solo/weaver/bin/solo-provisioner-daemon`. Nothing in the resolution path looks
there, so the path has to be re-supplied on every run.

**3. `profile == local` is a proxy for the wrong predicate.** The real condition is
"this build's version has no downloadable release". Local builds are not unstamped:
`Taskfile.yaml:17` defaults `VERSION` to `0.0.0`, and the daemon's catalog entry
(`pkg/software/infrastructure-catalog.yaml:315`) also defaults to `0.0.0`, so
auto-download resolves to a `v0.0.0` release tag that will never exist. The comments on
`FlagDaemonVersion` (`cmd/cli/commands/common/flags_common.go:323`) and in
`resolveDaemonBinarySource` still claim local builds report `"dev"`, which is stale.

## Proposed fix

Two changes that together remove the prompt from the local dev loop.

**Carry the daemon binary through CLI self-install.** `solo-provisioner install` is the
only command that opts out of `CheckWeaverInstallation` (`cmd/cli/commands/root.go:95`)
— every other command is forced to run as `/opt/solo/weaver/bin/solo-provisioner`. So
self-install is the single moment the CLI can still see the co-built daemon sibling in
the build tree's `bin/`. Add a best-effort step next to `InstallWeaver`
(`internal/workflows/steps/step_weaver.go:116`): if a daemon binary sits beside
`os.Executable()` (plain `solo-provisioner-daemon`, or suffixed
`solo-provisioner-daemon-<goos>-<goarch>`), install it to `paths.BinDir` using the same
temp-file + atomic-rename approach, after confirming it executes on this host. When no
sibling exists (a released install downloaded from GitHub), the step is a no-op.

This ties the daemon refresh to the moment a fresh build is actually in hand:
`task build && sudo bin/solo-provisioner-linux-<arch> install` always yields a matching
CLI + daemon pair, so there is no way to silently keep running a stale daemon.

**Replace the prompt with a deterministic resolution order.** `resolveDaemonBinarySource`
becomes:

1. `--daemon-bin` (explicit operator override, unchanged)
2. `SOLO_PROVISIONER_DAEMON_BIN` environment variable
3. an executable binary already at `paths.BinDir/solo-provisioner-daemon`
4. catalog auto-download, when this build's version resolves to a real release
5. prompt (interactive) or fail-fast error (non-interactive) only when all of the above miss

Step 3 must `stat` and platform-check the path directly rather than going through
`installer.IsInstalled()` — the base installer gates on a recorded manifest entry that a
self-install copy will not have written. With the self-install change in place, step 3
always hits in the local loop, so the prompt never appears.

Alongside that, drop `profile == local` as the trigger in favour of version
resolvability, and skip resolution entirely when the daemon service is already running
so problem 1 cannot recur.

### Rejected alternatives

**One binary, two entrypoints** (`ExecStart=… solo-provisioner daemon run`, or an
argv[0] multi-call binary). This would delete the problem permanently — no path, no
download, no CLI/daemon version skew ever — but it links every CLI dependency (helm,
the Kubernetes client, teleport, the TUI) into a long-lived root-resident service,
contradicting the deliberate constraint recorded at `cmd/daemon/main.go:151` that the
daemon binary imports nothing under `cmd/cli`.

**Falling back to the latest published release for unstamped builds.** For local work
the daemon under test is usually the one just built; silently installing a *released*
daemon instead is worse than being prompted. Reasonable later as an explicit
`--daemon-version=latest` convenience, not as a default.

## Acceptance

- [ ] `sudo bin/solo-provisioner-linux-<arch> install` run from a build tree also
      installs the co-built `solo-provisioner-daemon` to `/opt/solo/weaver/bin/`, and is
      a silent no-op when no sibling daemon binary is present.
- [ ] `block node install --profile=local --traffic-shaping-enabled` with a daemon
      binary already at `paths.BinDir` completes with **no** daemon-binary-path prompt.
- [ ] `SOLO_PROVISIONER_DAEMON_BIN=<path>` is honoured as a source, ahead of the
      already-installed binary and the catalog.
- [ ] Resolution is skipped when the daemon service is already running — no prompt is
      shown for a value that would be discarded.
- [ ] The non-interactive path still fails fast with actionable resolution hints when no
      source can be resolved.
- [ ] `--daemon-bin` continues to take precedence over every other source, unchanged.
- [ ] Stale `"dev"`-version comments on `FlagDaemonVersion` and in
      `resolveDaemonBinarySource` are corrected to reflect the `0.0.0` default.

### Related Issues

* Follow-up to #917 (daemon catalog entry) and #909 (install-time daemon orchestration).

brunodam:
## Description

Two related gaps in how the `solo-provisioner-daemon` binary gets onto a host and off it
again.

`block node install --profile=local` with traffic shaping enabled always stopped to prompt
for a daemon binary path. That prompt is replaced with a precedence-ordered resolution, and
CLI self-install now carries the co-built daemon binary onto the host so the resolution
finds it without asking anyone anything.

Self-install placing a daemon binary made an existing asymmetry worse: self-*uninstall*
removed only the CLI, its symlink, and the sudoers drop-in. The daemon binary, its systemd
service, the network boot units, and the config tree under `/etc/solo-provisioner` all
survived it. Uninstall now tears all of that down, behind a required `--yes`.

### Three defects sat behind the one prompt

1. Resolution ran during the prompt phase, but its consumer `ensureBlockNodeDaemon` returns
   early when the daemon service is already running — so in the normal dev loop (daemon up,
   re-installing the BN) the operator answered a question whose answer was then discarded.
2. Nothing looked at `paths.BinDir/solo-provisioner-daemon`, so a daemon binary already
   installed on the host did not count as a source.
3. `profile == local` was a proxy for the real predicate, "this build's version has no
   downloadable release". Local builds are stamped `0.0.0` by the Taskfile's `VERSION`
   default — not `"dev"`, as the comments on `FlagDaemonVersion` and
   `resolveDaemonBinarySource` claimed — and no `v0.0.0` release exists to download.

### The resolution order

`resolveDaemonBinarySource` walks, in descending precedence: `--daemon-bin`, a
running-daemon short-circuit, `SOLO_PROVISIONER_DAEMON_BIN`, an executable binary already
installed in the weaver bin directory, then a catalog download when the version resolves to
a published release.

**An exhausted list is an error, not a prompt.** A path to a locally-built binary is not a
deployment choice to make at the keyboard, and keeping it out of the TUI leaves one resolver
rather than a prompting and a non-prompting variant — no `--force`/`ShouldPrompt` gate, no
`ChosenValues` threading at the call sites, and `internal/ui/prompt` and `pkg/sanity` drop
out of the file entirely. `upgrade` can then share the resolver outright: nothing prompts,
so its silent-convergence contract holds by construction rather than via a second code path.

What makes the installed-binary source actually hit in the local loop: `solo-provisioner
install` is the only command that runs from outside the weaver bin directory — every other
command is pinned there by `CheckWeaverInstallation` — so it is the only point at which the
CLI can still see the daemon binary that `task build` emitted next to it. A new best-effort
self-install step copies that sibling to `paths.BinDir`, tying the daemon refresh to the
moment a fresh build is in hand.

### Incidental fix: a self-copy that truncated the installed binary

`copyBinaryFile` opens its destination `O_TRUNC`, so on `main`:

```bash
sudo solo-provisioner daemon service install \
  --daemon-bin /opt/solo/weaver/bin/solo-provisioner-daemon
```

zeroes the installed daemon binary. The new resolution order can legitimately resolve to
exactly that path, so a `sameFile` guard was required — but the bug is independently
reachable today with an explicit flag.

**The guard precedes the `--version` platform probe.** The probe's job is to reject a binary
that cannot run on this host *before it replaces a working one*; a source that already IS
the installed binary has nothing to replace and cleared the same probe when it was
installed. Re-probing only spends a subprocess on what the new resolution order makes the
most common path. The bin-directory ownership and mode hardening stays ahead of the guard —
hosts that already carry a daemon binary are exactly the older installs that may still have
the group-writable directory it corrects.

### Completing the uninstall teardown

An operator who uninstalled the block node and cluster, then cleared `/opt/solo` by hand,
was left with a daemon process still running from a deleted inode — systemd reports it
`active`, so the resolver correctly skipped as "already running" — and three unit symlinks
dangling into a directory that no longer existed.

`NewSelfUninstallWorkflow` now removes the daemon service and binary, the config tree under
`/etc/solo-provisioner`, and both network boot units. `RemoveNetworkConfig` runs *before*
`NftServiceTeardown`: that step keeps the shared nft unit while the host-firewall `.nft`
file is present, so deleting the files first is what lets it take the unit rather than
orphaning one.

Uninstall refuses while a Kubernetes cluster is still provisioned — every teardown command
lives in the binary about to be deleted. The check stats kubeadm's admin kubeconfig rather
than reading recorded state, which can disagree with the host after a partial teardown; a
block node cannot outlive its cluster, so one probe covers both.

The command requires `--yes` and errors without it, with **no `-y` short form** — the point
is an act of confirmation, and `-y` is the flag operators append by reflex. `--purge` would
have been the wrong name: in the apt/dpkg/snap lineage it modifies *scope* ("also delete
config"), implying a bare uninstall is valid and does less.

Live kernel state — loaded nft tables, tc qdiscs — is left alone. It does not survive a
reboot once the boot-replay units and their inputs are gone, and flushing it would put
packet handling at risk on a host the operator is still connected to.

### Scope note: `block node upgrade`

`upgrade` passed a zero `DaemonBinarySource`, so on a local build it would try to download a
`v0.0.0` release and fail. It now shares `resolveDaemonBinarySource` and propagates its
error instead of discarding it. Slightly beyond the filed issue; happy to split it out if
reviewers prefer.

### Files changed

| File | Change |
|---|---|
| `cmd/cli/commands/block/node/daemon_offer.go` | Single precedence-ordered `resolveDaemonBinarySource` returning `(source, error)`; adds `installedDaemonBinaryPath`, `isDownloadableDaemonVersion`, `unresolvedDaemonBinaryError`; drops the prompt, the `profile`/`args`/`cv` parameters, and the `prompt`/`sanity` imports |
| `cmd/cli/commands/block/node/install.go` | Call-site update for the reduced signature |
| `cmd/cli/commands/block/node/reconfigure.go` | Same |
| `cmd/cli/commands/block/node/upgrade.go` | Shares the one resolver; propagates its error |
| `cmd/cli/commands/common/flags_common.go` | `--daemon-bin` description reflects auto-resolution; stale `"dev"` version comments corrected |
| `cmd/cli/commands/self_install.go` | `--yes` gate on `uninstall`, with resolution hints listing what is removed |
| `internal/workflows/steps/step_daemon.go` | New `sameFile` helper; same-path guard hoisted ahead of the `--version` probe; BinDir hardening kept ahead of the guard |
| `internal/workflows/steps/step_weaver.go` | New `InstallColocatedDaemonBinary` step, `findColocatedDaemonBinary` and `copyBinaryAtomic` helpers |
| `internal/workflows/steps/step_weaver_purge.go` | New — `CheckNoProvisionedCluster`, `RemoveNetworkConfig` |
| `internal/workflows/weaver.go` | Wires the colocated-daemon step into self-install; rebuilds `NewSelfUninstallWorkflow` as a 9-step teardown |
| `docs/quickstart.md` | `--daemon-bin` row, traffic-shaper daemon note, and a rewritten Uninstall section with the `--yes` flag and teardown order |
| `cmd/cli/commands/block/node/daemon_offer_test.go` | New — precedence order, version resolvability, installed-binary detection, error-not-prompt |
| `internal/workflows/steps/step_weaver_daemon_test.go` | New — sibling discovery, `copyBinaryAtomic`, `sameFile` |
| `internal/workflows/steps/step_weaver_purge_test.go` | New — uninstall preflight and config removal |

### Review guide

**Code review checklist**

- **Self-copy safety.** `sameFile` must precede every `copyBinaryFile`/`copyBinaryAtomic`
  whose destination could equal its source. Two such sites: `step_daemon.go` (the resolved
  source may be the installed binary) and `step_weaver.go` (re-running `install` from
  `/opt/solo/weaver/bin` makes the search dir the destination dir). A miss here zeroes the
  installed daemon.
- **Guard placement.** The `sameFile` short-circuit now runs *before* the `--version` probe.
  Confirm you accept that an already-installed binary is not re-probed, and that the
  BinDir `chmod`/`chown` still precedes the short-circuit — moving it after would skip the
  `2775` privilege-escalation fix-up on exactly the hosts that need it.
- **Resolver never prompts.** `resolveDaemonBinarySource` is shared by `install`,
  `reconfigure`, and `upgrade`. Any future prompt added to it silently breaks upgrade's
  silent-convergence contract.
- **Running-service short-circuit** returns an empty `BinPath` with a nil error. That is
  only correct because `ensureBlockNodeDaemon` never installs when the service is up —
  confirm that invariant still holds.
- **Best-effort contract.** No branch of `InstallColocatedDaemonBinary` returns a failure
  report. An official install has no sibling to find and must not fail because of it.
- **Uninstall step order.** `RemoveNetworkConfig` before `NftServiceTeardown`, and the
  cluster check first so a refusal costs nothing.
- **`cmd.Flag` vs `cmd.Flags()`** in `isDownloadableDaemonVersion`: `SetVarP` registers on
  `PersistentFlags`, which `Flags()` only reflects after cobra's pre-Execute merge.
- **No new dependencies.**

**Test commands**

macOS cannot build these packages (`internal/mount` is Linux-only). In the VM:

```bash
task vm:test:unit
```

New cases: `TestFindColocatedDaemonBinary`, `TestSameFile`, `TestCopyBinaryAtomic`,
`TestInstallColocatedDaemonBinary_NoSiblingSucceedsWithoutInstalling`,
`TestIsDownloadableDaemonVersion`, `TestInstalledDaemonBinaryPath`,
`TestResolveDaemonBinarySource`, `TestCheckNoProvisionedCluster`,
`TestRemoveNetworkConfig`.

**Manual UAT**

1. *Self-install carries the daemon.* `task build`, then in the BN VM
   `sudo bin/solo-provisioner-linux-arm64 install`. Expect
   `/opt/solo/weaver/bin/solo-provisioner-daemon` present, mode 0755, mtime matching the
   build.
2. *No prompt on a local install.*
   `sudo solo-provisioner block node install --profile=local --traffic-shaping-enabled --egress-interface enp0s1`.
   Expect no "Daemon Binary Path" prompt and
   `INF Using the daemon binary already present on this host`, followed by
   `INF Daemon binary is already installed at the destination, skipping copy` — with no
   `--version` subprocess in between.
3. *Self-copy guard.* `sha256sum` the installed daemon, `daemon service stop`, then
   `daemon service install --daemon-bin /opt/solo/weaver/bin/solo-provisioner-daemon`, then
   `sha256sum` again. Expect an identical digest (not 0 bytes). On `main` this truncates.
4. *Re-running self-install is idempotent.* `sudo /opt/solo/weaver/bin/solo-provisioner install`
   twice; the daemon digest must not change.
5. *Env override.* `sudo SOLO_PROVISIONER_DAEMON_BIN=<path> solo-provisioner block node reconfigure --traffic-shaping-enabled=true`.
   Expect `INF Using the daemon binary from the environment`.
6. *Unresolvable source fails, never prompts.* Remove the installed daemon binary, then run
   a local install **interactively**. Expect the `no solo-provisioner-daemon binary could be
   resolved` error with its resolution hints, and no prompt.
7. *Uninstall refuses without `--yes`.* `sudo solo-provisioner uninstall` — expect
   `refusing to uninstall without --yes` and a hint listing what would be removed.
8. *Uninstall refuses while a cluster exists.* With `/etc/kubernetes/admin.conf` present,
   `sudo solo-provisioner uninstall --yes` — expect the ordered teardown hint.
9. *Full teardown.* After `block node uninstall` and `kube cluster uninstall`, run
   `sudo solo-provisioner uninstall --yes`. Expect gone: the daemon binary and service,
   `/etc/solo-provisioner`, `/usr/local/sbin/solo-provisioner-bandwidth-shaper.sh`, all
   three units under `/usr/lib/systemd/system`, and `systemctl status
   solo-provisioner-daemon` reporting no such unit.

**Risks / rollback**

- **`uninstall` is now destructive by design.** It removes files outside the weaver home
  (`/etc/solo-provisioner`, `/usr/local/sbin`, `/usr/lib/systemd/system`) via `os.RemoveAll`
  on a derived path. The derivation is `filepath.Dir(firewall.HostNftPath)` — worth a
  second pair of eyes, since a wrong constant there deletes the wrong tree.
- **`uninstall` gains a required flag.** Any script calling bare `solo-provisioner uninstall`
  now fails. Intentional, but it is a breaking CLI change.
- Self-install writes one extra file into `paths.BinDir`. Additive, and the step cannot fail
  the install.
- The resolution order can select an already-installed *stale* daemon where `main` would
  have forced the operator to supply a fresh path. Mitigated by the self-install refresh on
  every `task build` cycle; `--daemon-bin` still overrides everything.
- `daemon.yaml` under the weaver config dir is deliberately **not** removed by
  self-uninstall; `daemon service uninstall` owns that.
- Rollback is a plain revert — no on-disk state or config schema changes.

**Not done, deliberately**

- Flushing live nft tables and tc qdiscs on uninstall (see above).
- Merging the daemon into the CLI binary. It would delete the resolution problem
  permanently, but links every CLI dependency into a long-lived root service, contradicting
  the constraint recorded at `cmd/daemon/main.go`.
- `--daemon-version=latest` release resolution. For local work the daemon under test is
  usually the one just built, so silently installing a released daemon is worse than failing.
- Injecting the host paths into `CheckNoProvisionedCluster` / `RemoveNetworkConfig` as
  parameters. Both read absolute paths, so their unit tests can only exercise the
  not-present branches; making them meaningful needs a signature change.

### Related Issues

* Closes #973

### DIFF
--- a/cmd/cli/commands/block/node/daemon_offer.go
+++ b/cmd/cli/commands/block/node/daemon_offer.go
@@ -3,15 +3,18 @@
 package node
 
 import (
+	"os"
+	"path/filepath"
+
 	"github.com/automa-saga/logx"
 	"github.com/hashgraph/solo-weaver/cmd/cli/commands/common"
 	"github.com/hashgraph/solo-weaver/internal/daemon"
-	"github.com/hashgraph/solo-weaver/internal/ui/prompt"
 	"github.com/hashgraph/solo-weaver/internal/workflows"
 	workflowsteps "github.com/hashgraph/solo-weaver/internal/workflows/steps"
 	"github.com/hashgraph/solo-weaver/pkg/models"
 	pkgos "github.com/hashgraph/solo-weaver/pkg/os"
-	"github.com/hashgraph/solo-weaver/pkg/sanity"
+	"github.com/hashgraph/solo-weaver/pkg/semver"
+	"github.com/hashgraph/solo-weaver/pkg/software"
 	"github.com/joomcode/errorx"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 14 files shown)


alex-au (approved):
Thank you for making this change!

PR #988, opened 2026-08-07
refactor(network): rename golden nft fixtures to match post-#960 artifact names

### LINKED ISSUE #987: refactor(network): rename golden nft fixtures still carrying pre-#960 artifact names
## Problem

PR #960 renamed the two block-node nftables tables and their on-disk artifacts so the names describe what they do (`inet host` → `inet weaver-host-firewall`, `inet weaver` → `inet weaver-workload-policy`, and the matching `/etc/solo-provisioner/*.nft` files). It updated the golden fixtures' **contents** but not their **filenames**, so three fixtures still carry the pre-#960 names:

```
internal/network/firewall/testdata/network-host.golden.nft
internal/network/policy/testdata/network-weaver.golden.nft
internal/network/policy/testdata/network-weaver-dualstack.golden.nft
```

Every line inside `network-host.golden.nft` now reads `weaver-host-firewall`:

```
add table inet weaver-host-firewall
delete table inet weaver-host-firewall
add table inet weaver-host-firewall
table inet weaver-host-firewall {
```

That is exactly the filename/content mismatch #960 set out to remove. The fixture names are meant to mirror the rendered artifact names (`HostNftPath` at `internal/network/firewall/paths.go:27`, `WeaverNftPath` at `internal/network/policy/paths.go:20`), and that correspondence is what makes it obvious which fixture covers which `/etc/solo-provisioner/` file. Right now a reader grepping for `network-weaver-workload-policy` finds the renderer and the docs but not its own golden file.

## Proposed fix

Pure test-fixture rename — `git mv` plus the reference updates. No production code, template, or docs change.

| Before | After |
|---|---|
| `internal/network/firewall/testdata/network-host.golden.nft` | `internal/network/firewall/testdata/network-weaver-host-firewall.golden.nft` |
| `internal/network/policy/testdata/network-weaver.golden.nft` | `internal/network/policy/testdata/network-weaver-workload-policy.golden.nft` |
| `internal/network/policy/testdata/network-weaver-dualstack.golden.nft` | `internal/network/policy/testdata/network-weaver-workload-policy-dualstack.golden.nft` |

Four references to update, all in test code:

- `internal/network/firewall/firewall_test.go:365` — the regenerate-the-golden comment
- `internal/network/firewall/firewall_test.go:367` — `goldenPath`
- `internal/network/policy/policy_test.go:144` — `goldenPath`
- `internal/network/policy/policy_test.go:160` — dual-stack `goldenPath`

The new policy names are mildly redundant (`weaver…policy` inside `policy/testdata/`), but keeping the exact mirror of the artifact filename is worth more than the brevity of dropping the prefix.

## Acceptance

- [ ] All three golden fixtures are renamed, with the dual-stack variant included
- [ ] No file under `internal/network/` matches `network-host` or `network-weaver.` any more
- [ ] Each golden filename matches the artifact basename in `firewall/paths.go` and `policy/paths.go`
- [ ] `go test -tags='!integration' ./internal/network/firewall/... ./internal/network/policy/...` passes with no regenerated golden content — only the filenames change
- [ ] The renames are recorded as `git mv` (rename detection intact in the diff), not delete + add

alex-au:
## Description

#960 renamed the block-node nftables tables and their on-disk artifacts, and it updated the golden fixtures' **contents** — but not their **filenames**. The result was three fixtures whose names still read `network-host` / `network-weaver` while every line inside them read `weaver-host-firewall` / `weaver-workload-policy`, which is exactly the mismatch #960 set out to remove.

This renames the fixtures to mirror the rendered artifact basenames (`HostNftPath` in `internal/network/firewall/paths.go:27`, `WeaverNftPath` in `internal/network/policy/paths.go:20`), so the fixture covering each `/etc/solo-provisioner/*.nft` file is findable by grepping the artifact name. Golden content is byte-identical — the fixture changes are pure `git mv` (0 insertions, 0 deletions, 100% rename similarity).

No production behavior change.

### Renames

| Before | After |
|---|---|
| `internal/network/firewall/testdata/network-host.golden.nft` | `…/network-weaver-host-firewall.golden.nft` |
| `internal/network/policy/testdata/network-weaver.golden.nft` | `…/network-weaver-workload-policy.golden.nft` |
| `internal/network/policy/testdata/network-weaver-dualstack.golden.nft` | `…/network-weaver-workload-policy-dualstack.golden.nft` |

### Scope note: atomic-write temp prefixes

The issue covered only the three fixtures. The stale-name sweep also turned up the two `os.CreateTemp` prefixes used by the atomic writers, which carried the same pre-#960 names:

```go
// internal/network/firewall/render.go:114
tmp, err := os.CreateTemp(dir, ".network-host-*.nft.tmp")
// internal/network/policy/render.go:423
tmp, err := os.CreateTemp(dir, ".network-weaver-*.tmp")
```

These stage into `/etc/solo-provisioner/` beside the renamed artifacts, so a crash mid-write leaves an operator looking at `.network-host-1234.nft.tmp` next to `network-weaver-host-firewall.nft` — a name no artifact uses any more. Renamed to match. Nothing globs, stats, or sweeps on either prefix (`CreateTemp`'s return value is consumed directly for the fsync + rename), so this is a pure cosmetic change to a transient filename.

### Review guide

**Checklist**

- The three `.nft` fixture changes are recorded as renames with no content delta — verify with `git show --stat -M --find-renames=100% <sha> -- '*.nft'` (expect `0 insertions(+), 0 deletions(-)`).
- Each new fixture basename matches the artifact basename in `firewall/paths.go:27` and `policy/paths.go:20`.
- `internal/network/firewall/firewall_test.go:368` — `goldenPath` points at the renamed fixture; the regenerate comment above it (`:365-367`) names the same file.
- `internal/network/policy/policy_test.go:144` and `:160` — both `goldenPath` values updated, dual-stack variant included.
- `render.go` temp prefixes are `os.CreateTemp` patterns only — confirm no caller globs `.network-*` or depends on the old prefix.

**Tests**

```bash
# macOS-runnable (render + golden)
go test -tags='!integration' ./internal/network/firewall/... ./internal/network/policy/... ./internal/network/shape/...

# full unit suite (Linux-only packages) — in the UTM VM
task vm:test:unit
```

Result on this branch: 233 tests passed across the three packages; `task lint` reports 0 issues with no gofmt changes.

**Manual UAT**

Not applicable — no runtime artifact name changes, no operator-visible surface. The only runtime-observable difference is the basename of a transient `.tmp` file that exists for the duration of an atomic write.

**Risks / rollback**

Effectively none. If a golden test fails after this, it means the rename dropped or altered fixture bytes — check `git show -M` for a non-zero delta on the `.nft` files. Rollback is a straight revert of the single commit.

### Related Issues

* Closes #987


brunodam (approved):
Thanks for fixing this

Issue #670, opened 2026-06-11
feat(cli): implement eso operator install subcommand


brunodam:
Part of #377.

## Goal

Implement `solo-provisioner eso operator install` which installs the External Secrets Operator (ESO) Helm chart into the cluster. This is the prerequisite for all other `eso` subcommands.

## Acceptance criteria

- Running `solo-provisioner eso operator install` installs the `external-secrets/external-secrets` Helm chart.
- The command is idempotent: re-running it when ESO is already installed at the correct version is a no-op with a clear skip message.
- A `--namespace` flag controls the target namespace (default: `external-secrets`).
- A `--version` flag pins the chart version; the default is read from `infrastructure-catalog.yaml`.
- The ESO chart repo and version are pinned in `pkg/software/infrastructure-catalog.yaml`.

## Implementation notes

Follow the `step_alloy.go` / `step_metallb.go` pattern:
- Use `pkg/helm.Manager` (`hm.AddRepo`, `hm.IsInstalled`, `hm.InstallChart`) — not the `helm` binary directly.
- Place the command under `cmd/cli/commands/eso/operator/install.go`.
- Register the `eso` root command in `cmd/cli/commands/root.go` (create `cmd/cli/commands/eso/eso.go` as the parent cobra command, similar to `cmd/cli/commands/alloy/alloy.go`).
- Record installation state via the software state file (same pattern as other installers).

## Files to create / modify

| File | Change |
|------|--------|
| `cmd/cli/commands/eso/eso.go` | New — `esoCmd` parent command |
| `cmd/cli/commands/eso/operator/operator.go` | New — `operatorCmd` sub-group |
| `cmd/cli/commands/eso/operator/install.go` | New — install logic |
| `cmd/cli/commands/root.go` | Register `esoCmd` |
| `pkg/software/infrastructure-catalog.yaml` | Pin ESO chart version and repo URL |

## Tests

- Unit test: verify the Helm install call is made with the correct chart, repo, namespace, and version.
- Unit test: verify idempotency — when `IsInstalled` returns true, `InstallChart` is not called.

Issue #377, opened 2026-02-20
Create a family of solo-provisioner eso commands


brunodam:
The user, if they don't want to manage the k8s secret for the Alloy remotes manually, should be able to optionally use `solo-provisioner eso` commands to set up External Secret Operator in order to create the necessary k8s secrets for alloy.

PR #995, opened 2026-08-11
feat(cli): pin ESO version via infrastructure-catalog and remove --chart-version from eso operator install

### LINKED ISSUE #894: feat(cli): pin ESO version via infrastructure-catalog and remove --chart-version from eso operator install
Part of #377. Follow-up to #670 / PR #828.

## Problem

`eso operator install` (merged in #828) exposes a `--chart-version` flag that lets the caller pick the ESO chart version at install time. This gives operators more control than we want over a privileged, cluster-wide, cluster-singleton component:

- The ESO version is already pinned in `pkg/software/infrastructure-catalog.yaml` under `cluster:` (`external-secrets … default: "0.20.2"`). That `default:` is meant to be the single source of truth for which ESO version the platform installs.
- `--chart-version` undercuts that. It's checksum-gated (an *undeclared* version is rejected, so it's not "any version off the internet"), but it still lets an operator select a different **catalog-declared** version and diverge from the pinned `default:` on a per-install basis. As more versions are added under `versions:` over time, this becomes a real footgun — different clusters silently running different ESO versions.
- It's also **inconsistent** with the sibling cluster-infra installers. `metallb`, `cilium`, `metrics-server`, and `alloy cluster install` take **no** version flag — they resolve the catalog default via `chartSpec("<name>")` and that's it. Only `block node` (a data-plane component with its own upgrade story) exposes `--chart-version`. ESO copied the wrong precedent.

For an operator that holds cluster-wide secret-read RBAC, "everyone runs the one validated, pinned version" is the behavior we want; version selection should be a catalog edit in a solo-provisioner release, not a runtime flag.

## Proposed fix

Remove `--chart-version` from `eso operator install` and make the ESO version resolve **exclusively** from the catalog `default:` — matching metallb/cilium/alloy.

Touch points:

| File | Change |
|------|--------|
| `cmd/cli/commands/eso/operator/install.go` | Drop the `--chart-version` flag registration and `flagESOChartVersion` var; stop passing `Version` into the workflow |
| `cmd/cli/commands/common/flags_common.go` | Remove `FlagESOChartVersion()` |
| `internal/workflows/steps/step_external_secrets.go` | Drop `Version` from `ESOInstallOptions` (keep `Namespace`); resolve via the catalog default (either `resolveCatalogChartVersion(name, "")` or revert to `chartSpec("external-secrets")`) |
| `internal/workflows/steps/step_external_secrets_test.go` | Update `Test_SetupExternalSecrets_VersionResolution` — the "declared version accepted" / "undeclared rejected via flag" assertions no longer apply once the flag is gone |
| `docs/quickstart.md` | Remove the `--chart-version` example and the flag-table row in the ESO section |

Keep `--namespace`. `resolveCatalogChartVersion` can stay as an internal helper (it's harmless and still used with `""`), or be collapsed back into `resolveCatalogChart` if it has no remaining callers — reviewer's discretion.

## Acceptance

- [ ] `eso operator install --help` no longer lists `--chart-version`.
- [ ] The installed ESO version is always the catalog `default:` for `external-secrets`; changing it requires editing `infrastructure-catalog.yaml` (i.e. a solo-provisioner release), not a runtime flag.
- [ ] `--namespace` still works.
- [ ] `docs/quickstart.md` no longer documents `--chart-version` for ESO.
- [ ] Behavior matches metallb/cilium/alloy (catalog-default-only, no version flag).

Dosik13:
## Description

Removes `--chart-version` from `eso operator install`: the ESO chart version now resolves
exclusively from the infrastructure catalog's `default:` (`0.20.2`), matching how every other
cluster-infra installer works (metallb, cilium, metrics-server, alloy — none expose a version
flag). For a privileged, cluster-wide singleton, version selection belongs to a
solo-provisioner release (a catalog edit), not a runtime flag.

What it does:

* Drops the `--chart-version` flag, its definition (`FlagESOChartVersion`), and the
  version-threading through the install workflow.
* Simplifies `SetupExternalSecrets` to `(namespace string) *automa.WorkflowBuilder` — it was
  the ONLY `Setup*` in the steps package returning an error, and that error existed solely to
  report an undeclared user-supplied version. It now exactly mirrors
  `TeardownExternalSecrets(namespace string)` and the 15 peer installers, resolving via
  `chartSpec("external-secrets")`. (Slight deviation from the issue's touch-point table,
  which suggested keeping a one-field options struct — the plain signature is the stronger
  match for acceptance criterion "behavior matches metallb/cilium/alloy".)
* Collapses `resolveCatalogChartVersion` back into `resolveCatalogChart` (the issue offered
  this as reviewer's discretion). #670 added the two-arg helper solely to resolve a
  caller-supplied `--chart-version`; with that flag gone nothing passes a non-default version,
  so the helper had no caller. The now-orphaned unit test is removed with it.
* Adds a one-line breadcrumb to the install help and quickstart: "The chart version is
  pinned by the infrastructure catalog."

### Review guide

* **Why the signature change is in scope** — the error return and options struct existed only
  to carry the user version; with it gone, keeping either would leave ESO the lone outlier
  among the cluster-infra installers for no benefit. Install/Teardown are now symmetric.
* **Why `resolveCatalogChartVersion` is collapsed** — once `--chart-version` is removed,
  nothing calls the two-arg helper; it's orphaned. There's no current or foreseeable feature
  that needs to resolve a non-default version, so it's folded back into `resolveCatalogChart`,
  which still resolves the catalog default exactly as before (metallb/cilium/alloy, which
  share it, are unaffected). If some future feature ever needs a specific version, re-adding
  the parameter is trivial.
* **Block node keeps `--chart-version` deliberately** — data-plane component with an
  operator-driven version as persisted state and a dedicated `upgrade` command; the asymmetry
  is principled.

### Manual UAT

Run on a VM/host with a reachable cluster (requires `sudo`).

```bash
# 1. Flag surface: --chart-version is gone; help shows the catalog-pinned line
sudo solo-provisioner eso operator install --help
# -> no --chart-version flag; Long text contains "pinned by the infrastructure catalog"
sudo solo-provisioner eso operator install --chart-version 0.20.2
# -> Error: unknown flag: --chart-version (non-zero exit, nothing runs)

# 2. Fresh install resolves the catalog default (uninstall first if ESO is present)
sudo solo-provisioner eso operator uninstall
sudo solo-provisioner eso operator install
sudo KUBECONFIG=/etc/kubernetes/admin.conf helm list -n external-secrets
# -> CHART column shows external-secrets-0.20.2 (the catalog default)

# 3. Idempotency unchanged: re-run skips
sudo solo-provisioner eso operator install
# -> "already installed, skipping" path, exit 0

# 4. --namespace still works (R3)
sudo solo-provisioner eso operator uninstall
sudo solo-provisioner eso operator install --namespace my-eso
sudo KUBECONFIG=/etc/kubernetes/admin.conf helm list -n my-eso
# -> release present in my-eso at 0.20.2
sudo solo-provisioner eso operator uninstall --namespace my-eso   # cleanup
```

### Related Issues

* Closes #894
* Part of #377

### DIFF
--- a/cmd/cli/commands/common/flags_common.go
+++ b/cmd/cli/commands/common/flags_common.go
@@ -112,15 +112,6 @@ func FlagESONamespace() FlagDefinition[string] {
 	}
 }
 
-func FlagESOChartVersion() FlagDefinition[string] {
-	return FlagDefinition[string]{
-		Name:        "chart-version",
-		ShortName:   "",
-		Description: "External Secrets Operator chart version to install (must be declared in the infrastructure catalog; defaults to the catalog default)",
-		Default:     "",
-	}
-}
-
 func FlagESOSecretStore() FlagDefinition[string] {
 	return FlagDefinition[string]{
 		Name:        "store",
--- a/cmd/cli/commands/eso/operator/install.go
+++ b/cmd/cli/commands/eso/operator/install.go
@@ -10,40 +10,29 @@ import (
 	"github.com/hashgraph/solo-weaver/internal/workflows"
 )
 
-var flagESOChartVersion string
-
 var installCmd = &cobra.Command{
 	Use:   "install",
 	Short: "Install the External Secrets Operator",
 	Long: `Install the External Secrets Operator (ESO) Helm chart into the cluster.
 
 The command is idempotent: when ESO is already installed in the target
-namespace, installation is skipped with a clear message.
+namespace, installation is skipped with a clear message. The chart version is
+pinned by the infrastructure catalog.
 
 Examples:
-  # Install ESO with defaults (namespace: external-secrets, catalog default version)
+  # Install ESO with defaults (namespace: external-secrets)
   solo-provisioner eso operator install
 
   # Install into a custom namespace
-  solo-provisioner eso operator install --namespace my-eso
-
-  # Pin a specific catalog-declared chart version
-  solo-provisioner eso operator install --chart-version 0.20.2`,
+  solo-provisioner eso operator install --namespace my-eso`,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		l := logx.As()
 		l.Debug().
 			Strs("args", args).
 			Str("namespace", flagESONamespace).
-			Str("chartVersion", flagESOChartVersion).
 			Msg("Installing External Secrets Operator")
 
-		wb, err := workflows.NewESOInstallWorkflow(workflows.ESOInstallOptions{
-			Namespace: flagESONamespace,
-			Version:   flagESOChartVersion,
-		})
-		if err != nil {
-			return err
-		}
+		wb := workflows.NewESOInstallWorkflow(flagESONamespace)
 
 		if err := common.RunWorkflowBuilder(cmd.Context(), wb); err != nil {
 			return err
@@ -56,5 +45,4 @@ Examples:
 
 func init() {
 	common.FlagESONamespace().SetVar(installCmd, &flagESONamespace, false)
-	common.FlagESOChartVersion().SetVar(installCmd, &flagESOChartVersion, false)
 }
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -1028,17 +1028,14 @@ Manage the External Secrets Operator (ESO), which syncs secrets from external st
 
 #### Install External Secrets Operator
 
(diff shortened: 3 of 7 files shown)


brunodam (approved):
LGTM

PR #997, opened 2026-08-12
refactor(network/firewall): collapse the ICMP path-health accepts per family

### LINKED ISSUE #996: Collapse the ICMP path-health accepts in the host firewall into one rule per family
## Problem

`input_icmp_ipv4` and `input_icmp_ipv6` in `network-weaver-host-firewall.nft.tmpl` spell the
non-management ICMP policy as a run of one-type-per-rule accepts followed by a rate-limited
echo-request accept and an explicit drop:

```
icmp type destination-unreachable accept
icmp type time-exceeded accept
icmp type echo-request limit rate 10/second accept
icmp type echo-request drop
```

The IPv6 chain has the same shape with four path-health types instead of two, so between them
the two chains spend ten rules expressing "accept the path-health types, and accept echo-request
only while it is under budget".

Every one of those rules is evaluated per packet, and the repetition makes the one thing that
actually differs — that only echo-request is metered — easy to miss when reading.

## Proposed fix

Invert the limit to match the excess rather than the allowance, then accept the whole type set
in one rule:

```
icmp type echo-request limit rate over 10/second drop
icmp type { destination-unreachable, time-exceeded, echo-request } accept
```

`limit rate over` matches only when the rate is exceeded, so the first rule discards surplus
pings and the second accepts everything that survives. The explicit trailing drop becomes
redundant. Six rules become two in the IPv6 chain and four become two in the IPv4 chain.

The token bucket must stay scoped to echo-request. Folding the path-health types into a single
metered rule would put them through one shared bucket, so a ping flood would starve
`destination-unreachable`, `time-exceeded` and (on IPv6) `packet-too-big` — silently breaking
Path MTU Discovery and traceroute exactly when the host is under load. These are error signals
and must remain unconditional.

## Acceptance

- [ ] `input_icmp_ipv4` and `input_icmp_ipv6` each express the non-management policy in two rules
- [ ] The rate limit still applies to echo-request only; path-health types stay unmetered
- [ ] `@mgmt_addrs` / `@mgmt_addrs6` sources continue to bypass the limit entirely
- [ ] NDP and MLD handling in the IPv6 chain is untouched
- [ ] `internal/network/firewall/testdata/network-weaver-host-firewall.golden.nft` regenerated
- [ ] Verified against a real `nft`: the table loads, re-applies idempotently, and over-budget
      echo-request is dropped while path-health types still pass under the same load

brunodam:
## Description

* Collapses the non-management ICMP policy in `input_icmp_ipv4` and `input_icmp_ipv6` from one
  rule per admitted type into a single accept per family, by inverting the rate limit to match
  the excess instead of the allowance.

Both chains previously spelled the same idea across several rules — accept each path-health type,
accept `echo-request` while under budget, then explicitly drop the surplus:

```
icmp type destination-unreachable accept
icmp type time-exceeded accept
icmp type echo-request limit rate 10/second accept
icmp type echo-request drop
```

`limit rate over` matches only when the rate is exceeded, so discarding the surplus first lets one
accept cover everything that survives, and the trailing drop becomes redundant:

```
icmp type echo-request limit rate over 10/second drop
icmp type { destination-unreachable, time-exceeded, echo-request } accept
```

Four rules become two on IPv4; six become two on IPv6.

### The meter stays scoped to `echo-request` — deliberately

The obvious further collapse is to meter the whole type set in one rule. That would be a security
regression: all admitted types would share a single token bucket, so a ping flood would starve
`destination-unreachable`, `time-exceeded` and — on IPv6 — `packet-too-big`. Those are error
signals, and losing them silently breaks Path MTU Discovery and traceroute exactly when the host is
under load. IPv6 routers never fragment, so a dropped `packet-too-big` blackholes any flow whose
path MTU is smaller than the sender's.

They stay unconditional. Only `echo-request` reaches the meter.

### Unchanged

Management sources are accepted above the meter and so were never metered; that ordering is
untouched. The IPv6 NDP and MLD rules, the hop-limit 255 enforcement, and the `nd-redirect`
exclusion are all untouched. `ct state invalid drop` still runs first in both chains.

### Files changed

| File | Change |
|---|---|
| `internal/templates/files/network/network-weaver-host-firewall.nft.tmpl` | Both ICMP chains collapsed; comments explain why the meter stays scoped |
| `internal/network/firewall/testdata/network-weaver-host-firewall.golden.nft` | Regenerated |
| `internal/network/firewall/firewall_test.go` | Assertions updated; ordering test now pins that the over-budget drop precedes the accept it guards |

### Verified against a real kernel, not by inspection

The rendered table loads and re-applies idempotently under nftables. For the semantics, 60
`echo-request` packets were sent at roughly 200/second through the two-rule form, **once per
address family**, since the two chains are edited independently and a v4-only check would not
cover the ICMPv6 rewrite:

```
icmp   type echo-request limit rate over 10/second burst 5 packets counter packets 50 bytes 4200 drop
icmp   type { destination-unreachable, echo-request, time-exceeded } counter packets 10 bytes 840 accept

icmpv6 type echo-request limit rate over 10/second burst 5 packets counter packets 50 bytes 5200 drop
icmpv6 type { destination-unreachable, packet-too-big, time-exceeded, parameter-problem, echo-request } counter packets 10 bytes 1040 accept
```

Identical in both families: 50 matched the over-budget drop and exactly 10 reached the accept. The
meter binds to `echo-request` alone, and no other type can reach it because the drop rule does not
match them — which is what keeps `packet-too-big` out of the shared bucket on IPv6.

### Test plan

```bash
go test ./internal/network/...
```

255 tests pass. `task lint` reports 0 issues.

### Manual UAT

On a provisioned node, from a source **outside** `@mgmt_addrs`:

```bash
# Under budget: replies normally
ping -c 5 <node>

# Over budget: surplus is dropped, and the counter shows where
ping -f -c 200 <node>
sudo nft -a list chain inet weaver-host-firewall input_icmp_ipv4
```

The `limit rate over` rule's counter must advance while the accept below it advances at roughly the
configured rate. From an `@mgmt_addrs` source the flood must **not** be limited at all, since that
accept sits above the meter.

Path health must survive the flood — run this during the flood and it must still complete:

```bash
traceroute <node>
```

### Stacking note

#989 also modifies `network-weaver-host-firewall.nft.tmpl`, in different chains
(`prerouting_blocklist` and `output`). A textual conflict is unlikely, but both PRs regenerate the
same golden file, so whichever merges second will need its golden regenerated.

### Related Issues

* Closes #996

### DIFF
--- a/cmd/cli/commands/network/firewall/add.go
+++ b/cmd/cli/commands/network/firewall/add.go
@@ -3,31 +3,30 @@
 package firewall
 
 import (
-	"github.com/joomcode/errorx"
 	"github.com/spf13/cobra"
 )
 
 var addCmd = &cobra.Command{
 	Use:   "add",
-	Short: "Add a single --mgmt-cidr, --blocked-cidr, or --in-cluster-port",
+	Short: "Add CIDRs and/or ports to a rule (--name), merging with what is already there",
+	Long: "Add addresses and/or ports to one rule of the host firewall. --name selects the rule: a reserved block " +
+		"(mgmt, blocked, in_cluster) or a named allow rule. Adding is idempotent — an entry already present is " +
+		"left alone.\n\n" +
+		"The --mgmt-cidr, --blocked-cidr and --in-cluster-port flags are retained shorthands that name their " +
+		"reserved block implicitly.",
 	RunE: func(cmd *cobra.Command, _ []string) error {
-		mgr := newManager()
-		switch {
-		case cmd.Flags().Changed("mgmt-cidr"):
-			return mgr.AddMgmtCIDR(cmd.Context(), flagMgmtCIDR)
-		case cmd.Flags().Changed("blocked-cidr"):
-			return mgr.AddBlockedCIDR(cmd.Context(), flagBlockedCIDR)
-		case cmd.Flags().Changed("in-cluster-port"):
-			return mgr.AddPort(cmd.Context(), flagInClusterPort)
-		default:
-			return errorx.IllegalArgument.New("one of --mgmt-cidr, --blocked-cidr, or --in-cluster-port is required")
+		name, cidrs, ports, err := resolveTarget(cmd)
+		if err != nil {
+			return err
 		}
+		return newManager().Add(cmd.Context(), name, cidrs, ports)
 	},
 }
 
 func init() {
-	addCmd.Flags().StringVar(&flagMgmtCIDR, "mgmt-cidr", "", "A single management CIDR to add")
-	addCmd.Flags().StringVar(&flagBlockedCIDR, "blocked-cidr", "", "A single operator block-list CIDR to add (dropped inbound, outbound, and forwarded)")
-	addCmd.Flags().IntVar(&flagInClusterPort, "in-cluster-port", 0, "A single in-cluster host-service port to add")
+	registerTargetFlags(addCmd, "add")
+	addCmd.Flags().StringVar(&flagMgmtCIDR, "mgmt-cidr", "", "A single management CIDR to add (shorthand for --name mgmt --cidr)")
+	addCmd.Flags().StringVar(&flagBlockedCIDR, "blocked-cidr", "", "A single operator block-list CIDR to add (shorthand for --name blocked --cidr)")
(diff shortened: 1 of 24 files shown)


alex-au (dismissed):
LGTM!

alex-au (approved):
LGTM!

PR #999, opened 2026-08-12
feat(network/firewall): support named allow rules and a declarative config file

### LINKED ISSUE #998: feat(network/firewall): support named allow rules and a declarative config file
## Problem

The host firewall is a hardcoded special case of a much more general shape. It renders exactly
three bindings, each fixed at template level:

| Binding | Source | Ports |
|---|---|---|
| mgmt SSH | `@mgmt_addrs` | `--ssh-port` (a single scalar) |
| in-cluster | pod CIDR literal | `@in_cluster_ports` |
| block list | `@blocked_addrs` | all (three hooks) |

Verified gaps in `internal/network/firewall/table.go` — it contains no `udp`, no `interval` and
no `auto-merge`:

- **No UDP.** Cilium needs it (VXLAN 8472).
- **No port ranges.** `2379-2380`, `10256-10259`, `30000-32767` are inexpressible.
- **One management group only**, with a single scalar port.
- **No per-source ICMP echo.** ICMP is either "unrestricted from mgmt" or "rate-limited from the
  world"; there is no way to grant unmetered echo to a named set of sources.
- **`in_cluster` is TCP-only** and its ports cannot be scoped per protocol.

Why this matters now: weaver needs to be self-sufficient on operator-managed hardware, where no
external configuration management exists to supply a host firewall. It must be able to express a
complete ruleset including all management access. Real rulesets in use on managed bare metal
express roughly a dozen address groups against nine port groups in ~25 bindings, with both UDP
and port ranges — none of which weaver can currently represent.

## Proposed fix

Generalise to **named allow rules** over a small set of **reserved first-class blocks**, keeping
today's CLI working unchanged.

### Reserved blocks

Reserved where weaver derives the value *and* omission is dangerous:

| Block | Why reserved |
|---|---|
| `mgmt` | Lockout if wrong or absent; also feeds the ICMP chains |
| `blocked` | Renders on three hooks (`prerouting` raw, `input`, `output`), not one |
| `in_cluster` | Cluster breaks if absent, and the value is auto-detected from `.spec.podCIDR` |

`mgmt.ports` becomes a **list** (`--ssh-port` stays as sugar for a single-element list). That
removes a special case, makes mgmt render like any other rule, and allows the `{22, 1024}` shape
real rulesets use.

### Named allow rules

Everything else is an operator-authored record: source set x port set x protocol, always accept.
No action flag is needed, because the only deny is the reserved block list.

### Config file

```yaml
version: 1

mgmt:
  cidrs: ["192.168.68.0/24"]
  ports: ["22"]

blocked:
  cidrs: []

in_cluster:                                    # both fields optional
  cidrs: ["10.4.0.0/14"]                       # omitted -> auto-detected
  ports: ["4244", "6443", "7472", "10250"]     # omitted -> defaults

allow:
  - name: k8s-node
    cidrs: ["10.0.0.0/24"]
    ports: ["6443", "2379-2380", "10250", "10256-10259"]
    proto: tcp

  - name: cilium-vxlan
    cidrs: ["10.0.0.0/24"]
    ports: ["8472"]
    proto: udp

  - name: admin
    cidrs: ["203.0.113.5/32"]
    ports: ["22", "6443"]
    proto: tcp
    icmp_echo: true
```

Field rules: `name`, `cidrs` and `ports` are required on `allow` entries; `proto` defaults to
`tcp`; `icmp_echo` defaults to false. Ports are **strings** so ranges are expressible without a
mixed int/string list. `mgmt`, `blocked` and `in_cluster` are reserved names an `allow` entry
cannot take. `mgmt` rejects `proto` and `icmp_echo`; `blocked` rejects `ports`.

**Semantics**, which differ deliberately between the two kinds and need saying in the help text:

- `allow:` is declarative — an entry absent from the file is **deleted**.
- Reserved blocks absent from the file are **derived or defaulted**, never deleted. To disable
  one, empty its list (`in_cluster: {cidrs: []}` renders no rule, as `blocked: {cidrs: []}`
  already does).

### CLI

Structure (which rules exist, and their protocol) is file-only; membership (both lists) is
mutable from the CLI. Deliberate split: structure changes are reviewed, membership changes are
operational and sometimes urgent.

```
network firewall create  --mgmt-cidrs … --blocked-cidrs … [--ssh-port] [--in-cluster-ports] [--pod-cidr]
network firewall create  --from-file rules.yaml
network firewall add     --name <X> [--cidr …] [--port …]
network firewall remove  --name <X> [--cidr …] [--port …]
network firewall set     --name <X> [--cidrs …|--cidrs-file …] [--ports …]
network firewall show   [--name <X>] [--output yaml]
network firewall delete  --name <X> | --all
```

Reserved blocks are addressable by their well-known names, so `--name` is **required** on
`add`/`remove`/`set`/`delete` and there is only one mode:

```
network firewall add --name mgmt     --cidr 198.51.100.4/32
network firewall add --name blocked  --cidr 203.0.113.9/32
network firewall add --name k8s-node --cidr 10.0.0.5/32
```

Existing flags are retained as aliases (`--mgmt-cidr`, `--blocked-cidr`, `--in-cluster-port` and
the plural forms on `set`), so every current invocation keeps working and the interactive install
flow is unchanged. Net new surface is `--name` on the verbs that take a target, `--port`/`--ports`
generalised from `--in-cluster-port(s)`, `--from-file` on `create`, and `--output yaml` on `show`.

### Render notes

- An `icmp_echo` rule goes in `input_icmp_ipv4` / `input_icmp_ipv6` **immediately after the mgmt
  accept and above the rate meter**, so those sources get unmetered echo. Post-#997 the meter is
  inverted (`limit rate over … drop` before a single accept), so no trailing drop is needed.
  Extend the ordering test #997 added to pin that named-set accepts precede the meter.
- IPv6 ND/MLD and the hop-limit 255 enforcement stay structural and must not be removable by any
  rule — IPv6 is non-functional without them.
- `mgmt` keeps its full ICMP type list as a rendered constant, which is why `icmp_echo` is
  meaningless on it.
- Port sets need `flags interval` + `auto-merge` to carry ranges.

### Reuse rather than reimplement

`internal/network/policy` already solves most of this and should be factored for shared use, not
duplicated: registry persistence (one JSON per record), per-family set derivation and naming
(`V6SetName`, `PortsSetName`), element canonicalisation and diffing (`CanonicalizeElements`,
`DiffElements`), the `meta nfproto` family-split render pattern, and the shared `flock` under
`/run/solo-provisioner/network/`. The action vocabulary and tier ordering stay plane-specific.

## Acceptance

- [ ] `allow` rules render as `<family> saddr @<name> <proto> dport @<name>_ports accept` in the
      per-family input chains, for both IPv4 and IPv6.
- [ ] UDP rules render correctly and are covered by a golden test.
- [ ] Port ranges (`2379-2380`) render, with `flags interval` and `auto-merge` on the port set.
- [ ] `mgmt.ports` is a list; `--ssh-port` still works and produces a single-element list.
- [ ] `in_cluster` with both fields omitted auto-detects the pod CIDR and applies default ports.
- [ ] `in_cluster: {cidrs: []}` renders no in-cluster rule.
- [ ] `--name` is required on `add`/`remove`/`set`/`delete` and accepts `mgmt`, `blocked`,
      `in_cluster` and any `allow` name.
- [ ] `delete --name mgmt` is refused; `delete --all` requires confirmation.
- [ ] `blocked` rejects `--port`; `mgmt` rejects `proto` and `icmp_echo`; reserved names are
      rejected for `allow` entries.
- [ ] `icmp_echo` rules land above the rate meter, pinned by an ordering test.
- [ ] `create --from-file` applies a full config; an `allow` entry absent from the file is removed,
      and reserved blocks absent from the file are defaulted rather than removed.
- [ ] `show --output yaml` round-trips: its output re-applied via `create --from-file` is a no-op.
- [ ] Every pre-existing invocation of `network firewall` still behaves identically (regression
      test over the current flag set).
- [ ] Rendered rulesets load and re-apply idempotently against a real kernel, single-stack and
      dual-stack — string comparison alone cannot show that `nft` accepts the document.
- [ ] `docs/quickstart.md` updated (repo rule for CLI surface changes), plus
      `docs/dev/traffic-shaper.md`.

## Out of scope

- **Block-node service ports do not belong here.** That traffic is forwarded, not delivered
  locally, so `input` rules for it never match. Peer access to block-node ports is the workload
  policy plane's concern. Same likely applies to the NodePort range.
- **Confirm-or-roll-back apply.** Once weaver owns management access on hardware with no console,
  a bad render is a truck roll. Worth its own issue, and it should land before this is relied on
  as the only host firewall.
- **Migration of hosts currently managed externally.** Separate work; note that both `input`
  chains being default-drop means the effective policy is their intersection, so removing the
  external table can only widen access, never narrow it.
- **Cluster-derived rules.** Weaver installs the cluster and Cilium, so it could derive the
  `k8s-node` and `cilium-*` rules rather than having the operator author them. Worth doing after
  measuring which ports are actually needed — on a single-node host most of the multi-node k8s
  port set is likely inert.
- **Address-group sugar.** A `groups:` section with `cidrs_from:` to avoid repeating a CIDR list
  across rules. Deferred until the duplication actually hurts; the term `group` is reserved for it.

brunodam:
## Description

The host firewall rendered exactly three hardcoded bindings. `internal/network/firewall/table.go`
contained no `udp`, no `interval` and no `auto-merge`, so there was no UDP, no port ranges, one
management group on a single scalar port, and no way to grant unmetered ICMP echo to a named set of
sources. That is not enough to express a *complete* host ruleset, which weaver now has to do on
operator-managed hardware where no external configuration management supplies one.

This generalises the table to **named allow rules over three reserved blocks**. `mgmt`, `blocked` and
`in_cluster` stay first-class because weaver derives or defaults their content and omitting one is
dangerous — an empty mgmt list locks the operator out, an absent in-cluster list breaks the cluster,
and the block list renders on three hooks rather than one. Everything else is an operator-authored
source list x port list x protocol accept:

```
ip  saddr @k8s-node tcp dport @k8s-node_ports accept
ip6 saddr @admin6   tcp dport @admin_ports    accept
```

**Structure is file-only; membership is CLI-mutable.** Adding a rule is a reviewed change, while
unblocking an operator is sometimes urgent. The same YAML is the `--from-file` input, the persisted
state, and the output of `show --output yaml`, so the round-trip is exact by construction rather than
by test.

### Config file

```yaml
version: 1
mgmt:
  cidrs: ["192.168.68.0/24"]
  ports: ["22"]                                # now a list; --ssh-port is sugar for one element
blocked:
  cidrs: []
in_cluster:                                    # both fields optional
  cidrs: ["10.4.0.0/14"]                       # omitted -> auto-detected
  ports: ["4244", "6443", "7472", "10250"]     # omitted -> defaults
allow:
  - name: k8s-node
    cidrs: ["10.0.0.0/24"]
    ports: ["6443", "2379-2380", "10250", "10256-10259"]
  - name: cilium-vxlan
    cidrs: ["10.0.0.0/24"]
    ports: ["8472"]
    proto: udp
  - name: admin
    cidrs: ["203.0.113.5/32", "2001:db8:5e5::/64"]
    ports: ["22"]
    icmp_echo: true
```

### Deliberate decisions a reviewer should weigh

**`allow:` is declarative; the reserved blocks are required.** An `allow` entry absent from an applied
file is deleted. A reserved block absent from it is an error, not a silent default — the earlier
revision of this PR defaulted it, and for `mgmt` that default is an empty address list under the
input chain's default drop, i.e. exactly the lockout the asymmetry was meant to prevent. `cidrs` is
required inside `mgmt` and `blocked` for the same reason. Omitting `in_cluster.cidrs` still means
"auto-detect this node's pod CIDR" and `in_cluster: {cidrs: []}` still means "render no rule", which
is why nil-vs-empty on a decoded slice is load-bearing and has its own test. Raised by @brunodam in
review.

**The YAML config replaces the rendered ruleset as the source of truth for the mutating verbs.**
`auto-merge` is required for port ranges and also collapses adjacent entries, so the kernel can read
back differently from what was written — re-deriving intent from the ruleset would be lossy. `Parse`
stays as a fallback that recovers the *reserved blocks* from a pre-existing artifact (both the old and
new renderings), so a host that lost its config file still yields its management allowlist rather than
an error leaving the operator no way back in. Named allow rules are deliberately not
reverse-engineered from nft syntax: losing them is recoverable, losing management access is not.

**One state file, not one JSON per record** — a deviation from what #998 suggested. A change to the
management allowlist must be all-or-nothing, and a partial write across several files could leave a
host reachable by nobody. Happy to split it if you prefer the `policy` registry shape.

**`create --from-file` follows create-if-missing**, like every other `create` in this repo, so
re-applying an edited file needs `--force`. The existing warning names the flag explicitly, so this is
friction rather than silence — but it is the one call here I could see going the other way.

**`icmp_echo` renders above the rate meter.** Post-#997 the meter is `limit rate over … drop`, so an
accept placed below it would never be reached under a flood — exactly when an operator needs their own
ping to work. Pinned by an ordering test in both families.

**Reserved blocks keep their shipped set names** (`mgmt_addrs`, `in_cluster_ports`) while allow rules
use the bare name, matching the workload plane's `@bn-publisher`. All three derivations go through one
function, and `Table.Validate` rejects collisions — an allow rule named `mgmt_addrs` would claim the
mgmt block's set, and one named `k8s6` would claim the v6 set of `k8s`. nft accepts a duplicate set
declaration silently and merges the membership, so this has to be caught before render.

### Backward compatibility

Every pre-existing invocation behaves identically. The per-block flags are retained as shorthands that
name their reserved block implicitly, `set` still accepts several of them in one call (as one nft
transaction), and bare `delete` still means `--all`. `TestBackwardCompatibleInvocations` drives each
older form end to end.

One behaviour change worth flagging: `delete --all` now asks for confirmation in an **interactive**
session. Non-interactive callers are unaffected (`prompt.ShouldPrompt` is false), and `--force` skips it.

### Files changed

| File | Change |
|---|---|
| `internal/network/firewall/rule.go` | **New.** `Rule`, `Proto`, reserved names, port-spec parsing/ordering, derived set naming |
| `internal/network/firewall/config.go` | **New.** `FileConfig` schema, strict YAML load, table conversion both ways |
| `internal/network/firewall/table.go` | `Table` becomes three reserved `Rule`s plus `Allow`; name lookup, upsert/delete, collision check |
| `internal/network/firewall/render.go` | Per-rule flattening, family scoping |
| `internal/network/firewall/manager.go` | Config-file store, generic `Add`/`Remove`/`Set`/`SetMany`/`DeleteRule`/`Apply`/`Table` |
| `internal/network/firewall/parse.go` | Reserved-block recovery from both renderings |
| `internal/templates/files/.../network-weaver-host-firewall.nft.tmpl` | Ranged set decls, per-family allow rules, `icmp_echo` above the meter |
| `cmd/cli/commands/network/firewall/*.go` | `--name`, `--from-file`, `--output yaml`, `--all`, `--cidrs-file` |
| `internal/workflows/steps/step_network_firewall.go` | Carries named allow rules across a reconcile re-render |
| `docs/quickstart.md`, `docs/dev/traffic-shaper.md` | CLI surface and the design rationale |

### Review guide

Start with `rule.go` and `table.go` (the model), then the template, then `manager.go`'s `load`/
`applyAndPersist` ordering. The CLI is mechanical by comparison.

The bug worth checking I actually fixed: `step_network_firewall.go` force re-renders on
`block node reconfigure`, and `config.yaml` has no field for allow rules — so a reconfigure would have
wiped every operator-authored k8s/Cilium/admin rule while reporting success. The step now carries them
across. `TestNetworkFirewallCreate_PreservesNamedAllowRules` covers it.

**Real-kernel verification** — string tests cannot tell you whether `nft` accepts the document:

```bash
cd internal/network/firewall/testdata
docker run --rm --privileged -v "$PWD:/t:ro" debian:stable-slim sh -c '
  apt-get update -qq && apt-get install -y -qq nftables &&
  for f in network-weaver-host-firewall.golden.nft network-weaver-host-firewall-allow.golden.nft; do
    nft -c -f /t/$f && nft -f /t/$f && nft -f /t/$f; done &&
  nft list set inet weaver-host-firewall k8s-node_ports'
```

Both goldens load and re-apply idempotently on nftables 1.1.3, dual-stack, with the range preserved as
a single element:

```
elements = { 2379-2380, 6443, 10250, 10256-10259 }
```

The full unit suite passes on Linux (`go test ./internal/... ./pkg/... ./cmd/...` after `task mocks`).
The `network firewall` CLI package is Linux-only via `internal/mount`, so it does not run on macOS.

Manual UAT on the VM harness:

```bash
task network:all && task network:ip
sudo solo-provisioner network firewall create --from-file rules.yaml --force
sudo solo-provisioner network firewall show --output yaml | diff - rules.yaml   # expect no drift
sudo solo-provisioner network firewall add --name k8s-node --cidr 10.0.0.9/32
sudo nft list chain inet weaver-host-firewall input_ipv4
ping <host>   # from an admin CIDR, under a concurrent ping flood from elsewhere
```

### Stacking note

Based on `00996-collapse-icmp-path-health-accepts` (#997), not `main`: that PR inverts the ICMP rate
meter and rewrites the chains this one inserts into, and the `icmp_echo` placement only makes sense
against the inverted form. **Merge #997 first**, then retarget this to `main`.

### Out of scope

Called out in #998 and unchanged here: block-node service ports (that traffic is forwarded, so an
`input` rule for it never matches), confirm-or-roll-back apply, migration of externally-managed hosts,
cluster-derived k8s/Cilium rules, and `groups:` address sugar.

Of those, **confirm-or-roll-back apply is the one that should land before this is relied on as the only
host firewall** on hardware with no console — a bad render is otherwise a truck roll.

### Related Issues

* Closes #998

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -695,15 +695,15 @@ ICMP is a fixed, safe ruleset: full ICMP from the management allowlist, and from
 ```yaml
 version: 1
 
-mgmt:
-  cidrs: ["192.168.68.0/24"]
-  ports: ["22"]
+mgmt:                                          # required
+  cidrs: ["192.168.68.0/24"]                   # required
+  ports: ["22"]                                # omitted -> 22
 
-blocked:
-  cidrs: []
+blocked:                                       # required
+  cidrs: []                                    # required; [] means block nobody
 
-in_cluster:                                    # both fields optional
-  cidrs: ["10.4.0.0/14"]                       # omitted -> auto-detected
+in_cluster:                                    # required
+  cidrs: ["10.4.0.0/14"]                       # omitted -> auto-detected; [] -> no rule
   ports: ["4244", "6443", "7472", "10250"]     # omitted -> the defaults above
 
 allow:
@@ -735,10 +735,27 @@ sudo solo-provisioner network firewall create --from-file rules.yaml --force
 | `proto`     | no       | `tcp` (default) or `udp`. nft has no combined match, so a service on both is two rules |
 | `icmp_echo` | no       | Grants unmetered `echo-request`, rendered above the rate meter                |
 
-Two semantics differ deliberately between the record kinds:
+**The file is the whole table.** Nothing is inherited from the host's current firewall — only `add`/`remove`/`set` merge with what is already there. Two consequences:
 
 - **`allow:` is declarative** — a rule absent from the file is **deleted**.
-- **Reserved blocks absent from the file are derived or defaulted, never deleted**, so a partial file cannot silently drop management access. To disable one, give it an empty list (`in_cluster: {cidrs: []}`).
+- **All three reserved blocks are required**, as is `cidrs` inside `mgmt` and `blocked`. An omitted block would fall back to a weaver default the file never stated, and for `mgmt` that default is an empty allowlist under the default-drop policy — a lockout nobody wrote down. To render no rule for a block, state it with an empty list (`in_cluster: {cidrs: []}`); the block still cannot be removed.
+
+| Key                | Required | Omitted means                                                     |
+|--------------------|----------|-------------------------------------------------------------------|
+| `version`          | no       | the current schema version (`1`)                                   |
+| `mgmt`             | **yes**  | — (rejected)                                                       |
+| `blocked`          | **yes**  | — (rejected)                                                       |
+| `in_cluster`       | **yes**  | — (rejected)                                                       |
+| `mgmt.cidrs`       | **yes**  | — (rejected: no safe default exists)                               |
+| `mgmt.ports`       | no       | `22`                                                               |
+| `blocked.cidrs`    | **yes**  | — (rejected; write `[]` to block nobody)                           |
+| `in_cluster.cidrs` | no       | auto-detect this node's pod CIDR (`[]` renders no in-cluster rule) |
+| `in_cluster.ports` | no       | `4244,6443,7472,10250`                                             |
+| `allow`            | no       | no named allow rules — **and any that exist are deleted**          |
+
+`in_cluster.cidrs` is the one address list weaver can legitimately derive on its own, which is why it stays optional; its absence costs a rule rather than access to the host.
+
+The same rule applies to the persisted config at `/etc/solo-provisioner/network-weaver-host-firewall.yaml`: a truncated or hand-edited file is refused rather than loaded with a defaulted management allowlist. Re-run `create --from-file` to repair one.
 
 #### Modify a Rule's Addresses / Ports
 
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -645,11 +645,18 @@ sudo solo-provisioner kube cluster uninstall --continue-on-error
 
 ### Network Commands
 
-Manage node-level network state behind the traffic shaper. The `firewall` scope manages the node-agnostic `inet weaver-host-firewall` nftables table — the host's own SSH/management allowlist, ICMP policy, and in-cluster host-service ports. It is separate from the `inet weaver-workload-policy` workload plane and applies to every node type (block, consensus, mirror, relay).
+Manage node-level network state behind the traffic shaper. The `firewall` scope manages the node-agnostic `inet weaver-host-firewall` nftables table — the host's own management allowlist, ICMP policy, in-cluster host-service ports, and any number of named allow rules. It is separate from the `inet weaver-workload-policy` workload plane and applies to every node type (block, consensus, mirror, relay).
+
+The table holds two kinds of record:
+
+- **Three reserved blocks** — `mgmt` (management allowlist), `blocked` (operator block list, dropped on `prerouting`, `input` and `output`), and `in_cluster` (host-service ports reachable from the pod CIDR). They are first-class because weaver derives or defaults their content and omitting one is dangerous.
+- **Named allow rules** — an operator-authored source list x port list x protocol accept, for anything else the host must admit (Kubernetes control-plane ports, Cilium VXLAN, an admin jump host).
+
+Structure — which rules exist, and what protocol each matches — is declared in a config file. Membership — the addresses and ports inside a rule — is mutable straight from the CLI. That split is deliberate: adding a rule is a reviewed change, while unblocking an operator is sometimes urgent.
 
 #### Create the Host Firewall
 
-create-if-missing: if the `inet weaver-host-firewall` table already exists, the command makes no changes unless `--force` is passed (which re-renders from the flags). Every mutation applies to the live kernel in one atomic `nft -f` transaction and atomically rewrites `/etc/solo-provisioner/network-weaver-host-firewall.nft`.
+create-if-missing: if the `inet weaver-host-firewall` table already exists, the command makes no changes unless `--force` is passed (which re-renders from the flags or file). Every mutation applies to the live kernel in one atomic `nft -f` transaction and atomically rewrites both `/etc/solo-provisioner/network-weaver-host-firewall.nft` and the config it was rendered from, `/etc/solo-provisioner/network-weaver-host-firewall.yaml`.
 
 ```bash
 # Create with a management allowlist and the default in-cluster ports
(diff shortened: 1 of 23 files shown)


docs/quickstart.md
```diff
@@ -667,52 +674,136 @@ sudo solo-provisioner network firewall create --mgmt-cidrs 10.0.0.0/8,192.168.0.
...
+
+  - name: cilium-vxlan
+    cidrs: ["10.0.0.0/24"]
+    ports: ["8472"]
+    proto: udp
+
+  - name: admin
+    cidrs: ["203.0.113.5/32", "2001:db8:5e5::/64"]
+    ports: ["22"]
+    icmp_echo: true
```
brunodam, line 723: Make sure icmp_echo should be false by default.

docs/quickstart.md
```diff
@@ -667,52 +674,136 @@ sudo solo-provisioner network firewall create --mgmt-cidrs 10.0.0.0/8,192.168.0.
...
+> There is no `--service-ports`: BN ports live only in `network policy --ports`. That traffic is forwarded rather than delivered locally, so an `input` rule for it would never match.
+
+#### Declare Named Allow Rules
+
+`--from-file` is the only way to declare a named allow rule, and it renders the whole table:
+
+```yaml
+version: 1
+
+mgmt:
```
brunodam: when this yaml file is specified, mgmt, blocked and in_cluster should be mandatory. Is that the case?
In other words, when using the file, everything should be sourced from it instead of trying to merge existing values (which only applies to set)

docs/quickstart.md
```diff
@@ -667,52 +674,136 @@ sudo solo-provisioner network firewall create --mgmt-cidrs 10.0.0.0/8,192.168.0.
...
 #### Show / Delete the Host Firewall
 
 ```bash
 # Show the live inet weaver-host-firewall table
 sudo solo-provisioner network firewall show
 
-# Remove the table and /etc/solo-provisioner/network-weaver-host-firewall.nft
-sudo solo-provisioner network firewall delete
+# Show the declarative config the ruleset was rendered from
+sudo solo-provisioner network firewall show --output yaml
```
brunodam, line 802: is show displaying the same format that was passed via --from-file or the contents of the nftable? I believe it should be the format of the --from-file which would help when we want to export those values to apply them again later. Let's check if it's worth the complexity.

docs/quickstart.md
```diff
@@ -667,52 +674,136 @@ sudo solo-provisioner network firewall create --mgmt-cidrs 10.0.0.0/8,192.168.0.
...
+sudo solo-provisioner network firewall show --name k8s-node
+
+# Delete one named allow rule
+sudo solo-provisioner network firewall delete --name k8s-node
+
+# Remove the whole table and its on-disk artifacts
+sudo solo-provisioner network firewall delete --all
+```
+
+`show --output yaml` prints exactly the schema `create --from-file` accepts, so it round-trips:
```
brunodam, line 814: what if the --output is not provided? Does it print the nftable format? If not, should we have an output type for that?

docs/quickstart.md
alex-au: confirmed, it is false by default. `Rule.ICMPEcho` is a plain bool (rule.go:67) with `yaml:"icmp_echo,omitempty"`, so it is false unless an allow rule opts in

docs/quickstart.md
alex-au: confirmed `show --output yaml` already prints exactly the `--from-file schema`

docs/quickstart.md
alex-au: Default is nft (show.go:87), which prints the live kernel ruleset via `nft list table inet weaver-host-firewall`

docs/quickstart.md
alex-au: makes sense, I'll implement this feature to make "mgmt, blocked and in_cluster" mandatory

alex-au:
## Manual UAT on a local VM

Ran the host firewall end to end on a Debian 13 / arm64 UTM VM (nftables v1.1.3, kernel
6.12.57) against `42a52e6`, driving the real CLI, the real kernel and the real systemd loader
unit — not the fake runner the unit tests use.

**11 of 13 planned cases pass. 1 pre-existing bug found and filed as #1002.**

| # | Case | Result |
|---|---|---|
| 1 | Render every rule kind from a file | :white_check_mark: 6/6 |
| 2 | Reserved blocks are required | :white_check_mark: 9/9 |
| 3 | `show --output yaml` round-trips | :white_check_mark: 6/6 |
| 4 | `allow:` is declarative, reserved blocks are not | :white_check_mark: 4/4 |
| 5 | Membership verbs by `--name` | :white_check_mark: 11/11 |
| 6 | Backward compatibility | :white_check_mark: 7/7 |
| 7 | create-if-missing | :white_check_mark: 2/2 |
| 8 | Delete semantics | :white_check_mark: 8/8 |
| 9 | `icmp_echo` above the rate meter | skipped |
| 10 | Set-name collisions refused before render | :white_check_mark: 3/3 |
| 11 | `block node reconfigure` preserves allow rules | not run — needs a provisioned BN |
| 12 | Recovery when the config file is lost | :white_check_mark: 4/4 |
| 13 | Boot persistence | :white_check_mark: |

### What each case actually confirmed

**1 — every rule kind renders correctly.** One `admin` rule with a mixed v4/v6 CIDR list split
into `ip saddr @admin` *and* `ip6 saddr @admin6`; `cilium-vxlan` rendered `udp` while its
neighbours rendered `tcp`; `icmp_echo: true` produced echo-request accepts in both families. The
range survived the kernel round-trip as a single element:

```
set k8s-node_ports { type inet_service; flags interval; auto-merge;
  elements = { 2379-2380, 6443, 10250, 10256-10259 } }
```

Re-applying the identical file produced a byte-identical `nft list` (`diff` clean).

**2 — the required-block rule added in `42a52e6`.** Each of these is rejected with the live table
left untouched (verified by diffing `nft list` before and after): missing `mgmt`, missing
`blocked`, missing `in_cluster`, `mgmt:` with a null body, `mgmt` without `cidrs`, `blocked: {}`.
The deliberate exception still works — `in_cluster` present with `cidrs` omitted is accepted and
auto-detects, warning and omitting the rule when no cluster is reachable.

**3 — round-trip.** `show --output yaml` piped back through `create --from-file --force` produced
an identical ruleset. `show` with no `--output` prints the live nft; `show --name k8s-node` prints
just that rule and nothing about `mgmt`; `show --name nope` errors.

**8 — delete.** `--name` removes one allow rule and leaves the table up; `--name mgmt` is refused
as reserved; `--name X --all` is refused as mutually exclusive; `--all --force` removes the table
and both artifacts. *Not covered: the new interactive confirmation prompt — I only exercised the
`--force` and non-interactive paths.*

**12 — recovery.** Deleting the persisted YAML and running `show --output yaml` still returns the
management allowlist, recovered from the rendered `.nft` (named allow rules correctly absent, as
documented). A *truncated* config — as opposed to a missing one — now fails loudly rather than
loading with a defaulted `mgmt` block.

**13 — boot persistence.** Rebooted the VM; `nft list table inet weaver-host-firewall` hashed
identically before and after, all six allow-rule lines present, loader unit `active`.

### Bug found: #1002 (pre-existing, not introduced here)

Adding a CIDR already covered by an existing member of the same rule breaks the apply:

```console
$ sudo solo-provisioner network firewall add --name k8s-node --cidr 10.0.0.5/32
  Error: service solo-provisioner-network-nft.service start failed

# journalctl:
network-weaver-host-firewall.nft:24: Error: conflicting intervals specified
        set k8s-node { type ipv4_addr; flags interval; elements = { 10.0.0.0/24, 10.0.0.5/32 }; }
```

The address sets are declared `flags interval` **without** `auto-merge`; the port sets in the same
template have it, which is why `--port` never hits this and `--cidr` does. That predates this PR
(the address sets have looked like this since fd3a73b), so **#1002 is filed separately rather than
fixed here**.

Two things about it are worth a reviewer's attention anyway:

1. **This PR makes it reachable by following the docs.** The example in `docs/quickstart.md` and in
   this PR's own description is `add --name k8s-node --cidr 10.0.0.5/32` against a `k8s-node` rule
   holding `10.0.0.0/24` — precisely the failing input. Named allow rules also multiply the number
   of address sets an operator can `add --cidr` into.
2. **The persisted artifact is poisoned.** `applyAndPersist` writes the `.yaml` and `.nft` *before*
   applying, so a rejected ruleset still becomes the boot artifact. The live table is fine (the unit
   has no `ExecStop`, so nothing is flushed) — but `nft -c -f` on the persisted file then fails, and
   the host reboots with no weaver firewall at all. This is the part that turns a clear CLI error
   into a silent loss of the firewall.

Also worth knowing while testing: the repeated failed starts trip systemd's start-limit, after
which every `network firewall` mutation fails with the same opaque error until
`systemctl reset-failed solo-provisioner-network-nft.service`. My first UAT run reported 18
failures for this reason; re-running with a reset between cases showed one real bug and everything
else green.

### Why the existing tests missed it

The unit suite asserts on rendered strings, and `elements = { 10.0.0.0/24, 10.0.0.5/32 }` is a
valid string — only nftables' parser rejects it. The real-kernel check in this PR's description
loads the *goldens*, which are `create` output with disjoint CIDRs; reproducing this needs a
`create` followed by an `add`, which no golden captures. And no fixture anywhere uses an
overlapping CIDR pair.

### Still open

- **UAT-11** (`block node reconfigure` preserves named allow rules) is the one case needing a
  provisioned block node, so it did not run on this 3 CPU / 3 GB VM. It is the behaviour the
  `step_network_firewall.go` fix in this PR is *for*, and it is covered by
  `TestNetworkFirewallCreate_PreservesNamedAllowRules` at the unit level. Deciding how to cover it
  end to end.

alex-au:
## UAT-11 now covered end to end — the `reconfigure` fix works

Followed up the earlier UAT by actually provisioning a block node on the VM, so the one case that
needed a real cluster is no longer outstanding. `-p local` fits the 3 CPU / 3 GB VM (the profile
asks for 3 CPU / 1 GB), so no VM resizing was needed.

```bash
sudo solo-provisioner kube cluster install -p local --non-interactive     # 5m1s, node Ready, 12 pods Running
sudo solo-provisioner block node install -p local --non-interactive      # Successfully installed Hedera Block Node
sudo solo-provisioner network firewall create --from-file rules.yaml --force
sudo solo-provisioner block node reconfigure -p local --non-interactive \
  --firewall-enabled --mgmt-cidrs 192.168.50.0/24
```

**Result: all three named allow rules survived the force re-render.**

```
ip  saddr @admin        icmp   type echo-request accept
ip6 saddr @admin6       icmpv6 type echo-request accept
ip  saddr @admin        tcp dport @admin_ports        accept
ip  saddr @cilium-vxlan udp dport @cilium-vxlan_ports accept
ip  saddr @k8s-node     tcp dport @k8s-node_ports     accept
ip6 saddr @admin6       tcp dport @admin_ports        accept

k8s-node_ports  elements = { 2379-2380, 6443, 10250, 10256-10259 }
mgmt_addrs      elements = { 192.168.50.0/24 }
```

6/6 rule lines, port ranges intact, `mgmt` correctly re-rendered from the flag, and the persisted
config still lists all 3 allow rules. This is the behaviour `step_network_firewall.go:104-111` was
added for — before it, `reconfigure` re-rendered from `config.yaml` (which has no field for allow
rules) and would have dropped every one of them while reporting success.

**UAT status is now 12 of 13.** Only UAT-9 (the `icmp_echo`-under-flood functional check) is
unrun, by choice — it needs a second source host. The kernel-side half of that claim (accept
renders above the rate meter) is already pinned by the ordering test in both families.

### Second bug found on the way: #1003

The first reconfigure attempt — **without** `--firewall-enabled` — deleted the host firewall
outright:

```
INF Removing host firewall (inet weaver-host-firewall)   step_id=network-firewall-delete
DBG Persisted host-firewall configuration into runtime state  disabled=true managementCidrs=0
INF Successfully reconfigured Hedera Block Node
```

Table, both artifacts and the management allowlist gone, exit 0. The cause is that
`ResolveHostFirewallConfig` seeds the enable/disable decision from `MachineState.Firewall`
(deliberately, per `host_firewall.go:132-139`), but the standalone `network firewall` verbs never
write that state — so a firewall created with `network firewall create` is invisible to the seed
and gets torn down.

Filed separately as #1003 since it predates this PR. It is worth flagging here anyway, because
**this PR makes `create --from-file` the only way to declare named allow rules** — so the richer
the hand-authored ruleset an operator builds through the path this PR adds, the more a stray
`reconfigure` destroys, with no way to rebuild it from `config.yaml`.

The two findings are complementary: #1003 is the enable decision flipping silently, while this PR
fixes the allow rules being dropped whenever it stays enabled.

One correction to my earlier comment: I said the interactive reconfigure prompt is seeded from
`IsActive` ground truth. It is not — `IsActive`'s doc comment describes that, but its only caller
is the rollback guard in `step_network_firewall.go:119`. The actual seed is `MachineState.Firewall`
on both paths, which is why `--non-interactive` is not special here; an interactive run defaults
the same prompt to No.

alex-au (approved):
UAT tested and this PR itself looks great, found two existing defects and raised issue #1002 and #1003 accordingly.

Issue #964, opened 2026-08-04
feat(daemon/shaper): raise default statusz poll interval to 5 min and document startup-probe guarantee


brunodam:
## Problem

The traffic-shaper monitor's default statusz poll interval is **5 seconds**.  For a steady-state background daemon that re-checks nft membership every tick, 5 s is too aggressive:

- It generates one sudo exec + worker process spawn every 5 s even when nothing has changed.
- Operators who set a long poll interval (e.g. 1 h) had no documented guarantee that traffic shaping would not be delayed by a full interval after a daemon restart or host reboot.

### 1 — Raise the default poll interval to 5 minutes

Change `DefaultStatuszPollInterval` (and its local mirror `defaultStatuszPollInterval`) from `5 * time.Second` to `5 * time.Minute`.  Update all inline comments, the embedded `daemon.yaml` docstring, the `--statusz-poll-interval` flag description, and the developer docs (`traffic-shaper-statusz.md`, `traffic-shaper.md`) to reflect the new default.

### 2 — Fix `statuszForceResyncInterval` to be meaningful relative to the new default

`statuszForceResyncInterval` is currently `time.Minute`.  When the poll interval is 5 minutes, `5 min > 1 min` is always true, so the force-resync condition fires on every single tick — making the digest-delta optimisation dead code.  Raise `statuszForceResyncInterval` to **1 hour** so at the default cadence only one in twelve ticks is a forced apply.

### 3 — Document the startup-probe guarantee (no implementation needed — already coded)

The poll loop already reconciles **once immediately on entry** before the first ticker tick.  After a daemon restart or host reboot, nft set membership is rehydrated as fast as the BN statusz endpoint responds — not bounded by the poll interval.  If statusz is not yet up, `superviseResponsibility` retries with exponential back-off.

This guarantee holds when `base_url` is configured.  When using pod discovery only (no `base_url`), the entry reconcile skips until the pod watcher finds a ready BN pod; convergence is then bounded by up to one poll interval after discovery.

## Acceptance criteria

- [x] `DefaultStatuszPollInterval` is `5 * time.Minute`; `defaultStatuszPollInterval` mirrors it
- [x] `statuszForceResyncInterval` is `time.Hour` (> poll default so the optimisation has teeth)
- [x] All inline comments, `daemon.yaml` embedded example, and flag descriptions reference `5m` not `5s`
[...]

Issue #918, opened 2026-07-23
feat(daemon): discover the BN statusz endpoint from the block-node pod instead of a fixed base_url


brunodam:
## Problem

The block-node traffic-shaper daemon reconciles its nft policy membership from a
block node's statusz endpoints (`statusz/inbound-clients`,
`statusz/outbound-clients`). Today the daemon only polls a **single, statically
configured** base URL from `daemon.yaml`:

```yaml
components:
  block_node:
    statusz:
      base_url: http://127.0.0.1:8080
      poll_interval: 5s
```

`StatuszConfig` (`internal/daemon/config.go`) documents this as the
"local-fallback statusz source" — a mock statusz server in dev/test, or a
directly reachable BN statusz — and its own doc comment flags the gap:

> When nil or with an empty BaseURL, the monitor has no statusz source to poll
> (BN-pod statusz discovery is a future story) and the poll loop idles.

So when no `base_url` is configured the poll loop stays inert: the daemon cannot
discover the statusz endpoint of the block-node pod it is running alongside on its
own. An operator must hand-configure a reachable URL, and that URL is fixed for
the daemon's lifetime — it does not follow the BN pod across restarts,
rescheduling, or IP changes.

With the poll scheduler now wired (#914 / #916), the fixed-`base_url` path works
end to end, which makes the missing auto-discovery the next limiting factor for a
zero-config production deployment.

## Proposed fix

Teach the block-node component to **discover** the BN pod's statusz endpoint
rather than relying solely on a configured `base_url`. The daemon already holds a
BN-scoped Kubernetes client and watches BN pods in the orbit namespace (the
pod-lifecycle watcher in `internal/daemon/blocknode/pod_watcher.go` already
resolves per-pod state), so the building blocks for locating the pod and its
statusz port/address are largely in place.

Direction (to be refined during design):

- Resolve the statusz endpoint from the watched BN pod (pod IP + the BN statusz
  port, or an in-cluster Service) instead of a hard-coded host URL.
- Keep the configured `base_url` as an explicit override / local-fallback (its
  current dev/test role), so discovery is additive and does not break the mock
  statusz workflow used in UAT.
- Re-resolve on pod lifecycle events so the endpoint follows the pod across
  restarts/reschedules rather than pinning to a stale address.
- Decide the precedence when both a discovered endpoint and a configured
  `base_url` exist, and the behavior when discovery fails (idle vs. fall back to
  `base_url`).

[...]

Issue #750, opened 2026-06-25
Story 3.5 — Monitor health: degraded state, `/status`, and alerting


alex-au:
**Epic:** #736 — BN traffic-shaper monitor

Design: [v4 design](https://github.com/automa-saga/traffic-shaper/blob/main/docs/bn-qos-multiclass-priority-design-v4-nft-priority.md) §8.1, §10 risk 3.

Surface monitor health so a crashed/stuck monitor (which silently loses ingress shaping on the next pod restart) is detected.

**Acceptance criteria:**
- Surface monitor state (including `degraded`) via `GET /block_node/traffic_shaper/status`.
- Alert on the monitor's `degraded` state, not only on pod events: a crashed monitor leaves stale qdiscs until the next pod restart, at which point ingress shaping is lost indefinitely.
- Document the daemonkit back-off behaviour and the operator response.

alex-au:
Field evidence for this story from an ENV-1 traffic-shaper E2E run (`main` @ `fd3a73b`, BN chart `0.40.0-rc1`).

**The endpoint reported healthy while the feature was completely dead for 2 h 13 m.**

The `statusz-poll` responsibility faulted on *every* tick for 2 h 13 m — statusz was returning
404 because the deployed BN did not serve `/statusz/*` — with the responsibility pinned at its
5-minute back-off ceiling the whole time:

```
Aug 11 16:43:31 WRN traffic-shaper responsibility faulted — retrying after back-off
  error="… statusz statusz/inbound returned status 404"
  responsibility=statusz-poll retry_in=300000
Aug 11 16:48:31 WRN traffic-shaper responsibility faulted — retrying after back-off
  … (repeats every 5 min for 2 h 13 m)
```

Throughout, the status socket reported healthy:

```
$ sudo curl --unix-socket /opt/solo/weaver/daemon/daemon.sock \
    http://x/block_node/traffic_shaper/status
{"monitor":{"state":"running"}}
```

All `bn-*` nft sets were empty for the entire period.

**Why it reports `running`:** `TrafficShaperStatusResponse` (`internal/daemon/blocknode/handler.go:16-18`)
carries only `daemonkit.MonitorState` from the supervisor's StatusTracker. `Run` never returns while
`superviseResponsibility` is absorbing faults (`traffic_shaper_monitor.go:145-163`), so the goroutine is
genuinely "running" — the per-responsibility fault is invisible at this level. The handler comment already
[...]

PR #1001, opened 2026-08-12
fix(daemon/shaper): wake the statusz poll loop on endpoint discovery

### LINKED ISSUE #1000: fix(daemon/shaper): entry reconcile loses a startup race with the pod watcher, delaying nft membership by a full poll interval
## Problem

After `block node install --traffic-shaping-enabled`, the `inet weaver-workload-policy`
sets stay **empty for a full poll interval (5 min by default)** — even though the BN pod
is `ContainersReady` and `statusz` is serving the correct roster for that entire window.

Measured on `main` @ `fd3a73b`, BN chart `0.40.0-rc1`, `local` profile, single-node VM,
daemon left on pod discovery (no `base_url` — i.e. the default install path):

| Time | Event | Source |
|---|---|---|
| 23:08:36 | BN pod reaches `ContainersReady=True` | pod-condition poll (2 s) |
| 23:08:37 | `block node install` workflow completes | `setup_report_20260812_230837.yaml` |
| 23:08:38 | daemon install workflow completes (802 ms) → `solo-provisioner-daemon` starts | `setup_report_20260812_230838.yaml` |
| **23:13:39** | `bn-publisher` / `bn-partner-out` / `bn-backfill` membership finally appears | nft poll (2 s) |

**Δ = 5 min 01 s — exactly one poll interval.**

Critically, the pod was ready **two seconds before the daemon started**, so this is not
"the daemon waited for the pod." Everything the daemon needed was in place at `T₀`.

`statusz` answered 200 with the correct roster throughout the window:

```
$ curl -s -o /dev/null -w 'inbound=%{http_code}\n' http://10.4.0.26:40983/statusz/inbound
inbound=200

$ curl -s http://10.4.0.26:40983/statusz/inbound | head -c 200
{
  "activeEndpoints": [{
    "local":  { "address": "192.168.50.22",  "port": "40984" },
    "remote": { "address": "192.168.50.201", "port": "*" },
    "category": "publisher",
```

…while nft stayed empty for five minutes:

```
$ sudo nft list set inet weaver-workload-policy bn-publisher
table inet weaver-workload-policy {
        set bn-publisher {
                type ipv4_addr
                flags interval
        }
}
```

…then populated correctly, all at once, at 23:13:39:

```
$ sudo nft list set inet weaver-workload-policy bn-publisher
                elements = { 192.168.50.201 }
$ sudo nft list set inet weaver-workload-policy bn-partner-out
                elements = { 192.168.50.202 }
$ sudo nft list set inet weaver-workload-policy bn-backfill
                elements = { 192.168.50.204 . 50980 }
```

### Root cause

The entry reconcile in `runStatuszPoll` loses a startup race with the pod watcher.

1. The daemon starts and `runStatuszPoll` immediately calls `runReconcile()`
   (`internal/daemon/blocknode/traffic_shaper_monitor.go:343`).
2. `effectiveStatuszURL()` reads `discoveredStatuszURL`, which the pod watcher has not
   populated yet — its initial list/watch sync has not completed.
3. The empty-URL branch returns `nil`, **not an error**
   (`traffic_shaper_monitor.go:276-278`). Because it is not an error,
   `superviseResponsibility` does not retry it with the 5 s→5 min back-off.
4. The loop then parks on `select { case <-ctx.Done(); case <-ticker.C }`
   (`traffic_shaper_monitor.go:347-355`). **`ticker.C` is the only wake-up source** —
   nothing signals the loop when the pod watcher later discovers the URL.

So a few milliseconds of startup race cost a full poll interval of convergence. The pod
watcher discovers the endpoint seconds later and the loop never finds out.

### Impact

Reading the `forward` chain with empty sets, a new inbound connection from a publisher to
the pod on `:40984` matches no accept rule — `@bn-publisher` and `@bn-publisher_ports` are
both empty, `ip saddr <podCIDR> accept` does not match (the source is the peer, not a pod),
and `ct state established,related` does not match a new flow — so it falls through to the
chain's terminal `drop`.

If that reading is right, the window is not merely *unshaped*, it is a **~5-minute inbound
outage for the block node** after every install, reconfigure, reboot, or daemon restart.
This is invisible from the node itself: host-originated traffic (e.g. a local `curl` to
statusz) takes OUTPUT, not FORWARD, so it keeps working throughout. Needs empirical
confirmation from an off-node source before we treat it as an availability bug rather than
a latency bug — see Acceptance.

### Relationship to #964

#964 §3 documented this as a known, accepted caveat:

> This guarantee holds when `base_url` is configured. When using pod discovery only
> (no `base_url`), the entry reconcile skips until the pod watcher finds a ready BN pod;
> convergence is then bounded by up to one poll interval after discovery.

Two things make that caveat worse in practice than it reads:

1. **The default install path is the pod-discovery path.** `block node install` does not
   set `base_url`, so the configuration where the guarantee does *not* hold is the one
   every operator gets by default. Raising the default 5 s → 5 min turned a 5-second gap
   into a 5-minute gap precisely there.
2. **The caveat understates the trigger.** It describes waiting for the pod to *become*
   ready. Here the pod was ready *before* the daemon started and it still took a full
   interval — the race is against the watcher's initial cache sync, not against pod
   readiness.

Note #990 (persist statusz-derived membership across reboot) would mitigate the reboot
case but not this one: on a fresh install there is no previous membership to restore.

### Why this surfaced now

This is a **composition regression** — two independently reasonable changes, five days apart:

| Date | Change | Effect on this bug |
|---|---|---|
| 2026-07-30 | #936 (issue #918) — discover the statusz endpoint from the BN pod instead of a fixed `base_url`. Enabled by the BN-side fix in hiero-ledger/hiero-block-node#3316. | Moves every default install **off** the `base_url` path (where the entry-reconcile guarantee holds) **onto** the pod-discovery path (where it does not). Latent cost at the time: ~5 s. |
| 2026-08-04 | #965 (issue #964) — default poll interval 5 s → 5 min. | The same latent race now costs **5 minutes** instead of 5 seconds — a 60× amplification. |

Before #936, operators set `base_url` explicitly in the `block node install` TUI — originally
a workaround for the BN-side bug fixed by hiero-ledger/hiero-block-node#3316. On that path
`effectiveStatuszURL()` returns the configured URL immediately
(`traffic_shaper_monitor.go:205-206`), the pod watcher is never consulted, and the entry
reconcile converges promptly. That is why the guarantee held in practice and this race went
unnoticed: essentially nobody was running the discovery path.

#964 was closed the same week and reasoned about the startup-probe guarantee independently
of the fact that the default had, five days earlier, moved onto the one path where §3 says
the guarantee does not apply.

**This is not an argument for reverting.** Pod discovery is the better behaviour — it follows
the pod across restarts, reschedules and IP changes, which a fixed `base_url` cannot. The fix
is to give the discovery path the same startup convergence the `base_url` path already has.
`--daemon-config` (#937) can still set `base_url` as an operator workaround, but it trades
away pod-following to get it.

## Proposed fix

Two changes, both in `internal/daemon/blocknode/`:

1. **Wake the poll loop on endpoint discovery.** Add a buffered `chan struct{}` signalled
   by the pod watcher whenever `discoveredStatuszURL` changes, and select on it alongside
   the ticker:

   ```go
   case <-ticker.C:
   case <-m.urlChanged:
   ```

   The existing digest gate already does the right thing here: a `"" → url` transition
   resets `lastDigest`/`lastApply`, so the triggered reconcile performs a fresh apply
   rather than short-circuiting on a stale digest.

2. **Make the entry reconcile wait for the watcher's initial cache sync** before concluding
   there is no URL (e.g. `WaitForCacheSync` with a bounded timeout). Without this, a fresh
   install always loses the race, and fix 1 alone leaves convergence dependent on watcher
   timing rather than guaranteeing it.

With both, convergence after install/restart is bounded by BN startup time — which is what
#964 §3 claimed — on the pod-discovery path as well as with `base_url`.

## Acceptance

- [ ] Poll loop wakes on statusz endpoint discovery, not only on `ticker.C`
- [ ] Entry reconcile waits for the pod watcher's initial cache sync before treating the URL as absent
- [ ] Fresh `block node install --traffic-shaping-enabled` (pod discovery, no `base_url`) populates `bn-*` set membership within seconds of the BN pod becoming ready, not one poll interval
- [ ] Same holds for `systemctl restart solo-provisioner-daemon` while a ready BN pod already exists
- [ ] Same holds after a host reboot
- [ ] Unit test: URL transitioning `"" → url` between ticks triggers a reconcile without waiting for the next tick
- [ ] Confirm (or refute) the inbound-drop impact by reaching `:40984` on the BN pod **from an off-node source** during the window; if confirmed, re-label as an availability bug
- [ ] `traffic-shaper-statusz.md` caveat and the `traffic_shaper_monitor.go:338-342` comment updated to match the new behaviour
- [ ] `task lint` passes

### Reproduction

Deterministic loop, ~2 min per iteration, with the plane already installed and converged:

```bash
# 1. remove the ready pod so no statusz endpoint exists
kubectl -n block-node scale statefulset <STS> --replicas=0
kubectl -n block-node wait --for=delete pod -l app.kubernetes.io/name=block-node-server --timeout=120s

# 2. restart the daemon with NO ready pod → entry reconcile silently no-ops
sudo systemctl restart solo-provisioner-daemon; echo "T0=$(date +%H:%M:%S)"

# 3. clear membership so repopulation is observable
sudo nft flush set inet weaver-workload-policy bn-publisher
sudo nft flush set inet weaver-workload-policy bn-partner-out

# 4. bring the BN back
kubectl -n block-node scale statefulset <STS> --replicas=1
```

Observer:

```bash
while :; do
  R=$(kubectl -n block-node get pod -l app.kubernetes.io/name=block-node-server \
        -o jsonpath='{.items[0].status.conditions[?(@.type=="ContainersReady")].status}' 2>/dev/null)
  EL=$(sudo nft list set inet weaver-workload-policy bn-publisher 2>/dev/null \
        | tr -d '\n' | sed -n 's/.*elements = {\(.*\)}.*/\1/p')
  printf '%s ready=%-5s publisher={%s}\n' "$(date +%H:%M:%S)" "${R:-none}" "${EL:-EMPTY}"
  sleep 2
done
```

Membership appears at `T₀ + poll_interval` regardless of when the pod became ready. Setting
`components.block_node.statusz.poll_interval` to `30s` and repeating shows convergence
tracking the interval rather than pod readiness — which is the diagnostic that separates
this from ordinary startup latency.

alex-au:
## Description

After `block node install --traffic-shaping-enabled`, the `inet weaver-workload-policy` sets stayed
empty for a **full poll interval (5 min default)** — even though the BN pod was `ContainersReady` and
statusz was serving the correct roster the whole time. Same on `reconfigure`, host reboot, and
`systemctl restart solo-provisioner-daemon`.

Measured on a local ENV-2 install **before** this change:

| Time | Event | Source |
|---|---|---|
| 23:08:36 | BN pod reaches `ContainersReady=True` | pod-condition poll (2 s) |
| 23:08:37 | `block node install` workflow completes | `setup_report_20260812_230837.yaml` |
| 23:08:38 | daemon install completes (802 ms) → daemon starts | `setup_report_20260812_230838.yaml` |
| **23:13:39** | `bn-publisher` / `bn-partner-out` / `bn-backfill` populate | nft poll (2 s) |

**Δ = 5 min 01 s — exactly one poll interval.** Note the pod was ready *two seconds before the daemon
started*, so this was never "waiting for the pod": everything the daemon needed was in place at `T₀`,
and statusz answered `200` with the right roster throughout.

### Root cause

`Run` starts the pod watcher and the poll loop concurrently (`traffic_shaper_monitor.go:126-136`), so
on a cold start the entry reconcile (`:343`) normally runs *before* the watcher's initial
`pods.List()` (`pod_watcher.go:72`) has recorded an endpoint. Then:

1. `effectiveStatuszURL()` returns `""`.
2. The empty-URL branch returns `nil`, **not an error** — so `superviseResponsibility` does not retry
   it with the 5 s→5 min back-off.
3. The loop parks on `select { <-ctx.Done(); <-ticker.C }`. **`ticker.C` was the only wake-up source**,
   so nothing told it the endpoint appeared moments later.

A few milliseconds of startup race cost a full interval of convergence.

### Why the buffer is load-bearing — the main thing to review

`urlChanged` is `make(chan struct{}, 1)` with a **non-blocking** send. That is not incidental: it is
what makes the signal order-independent. The watcher frequently signals *before* the poll loop reaches
its `select`. With capacity 1 the signal waits in the buffer and the `select` returns immediately; an
**unbuffered** send would find no receiver, drop silently, and reintroduce the exact race this closes —
just moved a few lines. `TestSignalURLChanged_RetainsSignalForLateReceiver` pins this property
directly, with no timing dependence.

### Relationship to #964 — this is a composition regression

#964 §3 documented the caveat as accepted: *"When using pod discovery only (no `base_url`) … convergence
is then bounded by up to one poll interval after discovery."* Two changes five days apart made that bite:

| Date | Change | Effect |
|---|---|---|
| 2026-07-30 | #936 (issue #918) — discover the endpoint from the BN pod instead of a fixed `base_url` (enabled by hiero-ledger/hiero-block-node#3316) | Moved every default install **off** the `base_url` path onto the pod-discovery path. Cost then: ~5 s |
| 2026-08-04 | #965 (issue #964) — default poll interval 5 s → 5 min | Same latent race now costs **5 min** — 60× |

Before #936, operators set `base_url` explicitly in the install TUI, which returns immediately from
`effectiveStatuszURL()` and never consults the watcher — which is why the guarantee held in practice and
this race went unnoticed. **Deliberately not reverting to `base_url`:** discovery follows the pod across
restarts, reschedules and IP changes, which a fixed URL cannot. This PR gives the discovery path the
startup convergence `base_url` already had.

The caveat also *understated* the trigger — it describes waiting for the pod to *become* ready, but here
the pod was ready before the daemon started. The race is against the watcher's initial list, not pod
readiness.

### Files changed

| File | Change |
|---|---|
| `internal/daemon/blocknode/traffic_shaper_monitor.go` | Add `urlChanged` (cap 1) + `signalURLChanged()`; `select` on it alongside `ticker.C`; rewrite the entry-reconcile caveat comment |
| `internal/daemon/blocknode/pod_watcher.go` | Signal on endpoint discovery/change (`recordDiscoveredStatusz`) and on clear-by-delete (`handlePodDelete`) |
| `internal/daemon/blocknode/traffic_shaper_monitor_test.go` | 4 tests: buffering guarantee, coalescing/nil-safety, signal-wakes-before-tick, loss signal is harmless |
| `internal/daemon/blocknode/pod_watcher_test.go` | 2 tests: signals only on change, signals when the endpoint is cleared |
| `docs/dev/daemon/traffic-shaper-statusz.md` | Rewrite the pod-discovery paragraph — convergence no longer bounded by `poll_interval` |
| `docs/dev/traffic-shaper.md` | Reboot table no longer says "pod readiness + up to one poll interval" |

No informer is involved — `runPodWatcher` is a raw `List` + `Watch`, so the `WaitForCacheSync` wording in
issue #1000's proposed fix was not implementable as written. It also turned out unnecessary: with the
buffered channel the entry reconcile does not need to observe a URL at all.

### Review guide

**Checklist**

- [ ] `traffic_shaper_monitor.go` — `urlChanged` is created with capacity **1**; a change to `0` silently
      restores the bug (only the new unit test would catch it)
- [ ] `signalURLChanged()` uses `select { case ch <- struct{}{}: default: }` — must never block, since it
      is called from the watch-event path
- [ ] Nil-channel guard is present: unit-test scaffolding constructs `TrafficShaperMonitor` directly, and a
      nil channel in `select` blocks forever rather than panicking (so a missing guard would be silent)
- [ ] `pod_watcher.go:279` — signal is sent **after** `m.mu.Unlock()`, never while holding the mutex
- [ ] `handlePodDelete` — signal is sent after the unlock **and before** the `if !ok { return }` early
      return, so a pod with no recorded veth still reports the loss
- [ ] `handlePodDelete` only signals when `endpointCleared` — an unrelated pod's delete must stay silent
- [ ] No change to the digest gate: a `"" → url` transition already resets `lastDigest`/`lastApply`
      (`:256-275`), so the woken reconcile does a fresh apply rather than short-circuiting

**Test commands**

```bash
go test -race -count=1 ./internal/daemon/blocknode/...
go test -race -count=1 -run 'TestSignalURLChanged|TestRunStatuszPoll_DiscoverySignal|TestRecordDiscoveredStatusz_SignalsOnlyOnChange|TestHandlePodDelete_SignalsWhenEndpointCleared' ./internal/daemon/blocknode/
task vm:test:unit    # full suite; internal/mount is Linux-only so macOS cannot run it
task lint
```

`TestRunStatuszPoll_DiscoverySignalWakesLoopBeforeNextTick` uses a **1-hour** poll interval, so the ticker
cannot mask a missing signal. Verified it fails without the fix — replacing the new `select` case with a
never-ready channel produces `timed out waiting for count >= 1 (last: 0)`.

### Manual UAT

Requires a BN host with the plane installed and converged, and a BN whose statusz serves
`/statusz/inbound|outbound` (verify with `curl -s -o /dev/null -w '%{http_code}\n' http://<podIP>:40983/statusz/inbound`
→ `200`). Run the observer in one shell throughout:

```bash
sudo -v
while :; do
  R=$(kubectl -n block-node get pod -l app.kubernetes.io/name=block-node-server \
        -o jsonpath='{.items[0].status.conditions[?(@.type=="ContainersReady")].status}' 2>/dev/null)
  EL=$(sudo nft list set inet weaver-workload-policy bn-publisher 2>/dev/null \
        | tr -d '\n' | sed -n 's/.*elements = {\(.*\)}.*/\1/p')
  printf '%s ready=%-5s publisher={%s}\n' "$(date +%H:%M:%S)" "${R:-none}" "${EL:-EMPTY}"
  sleep 2
done
```

Set `STS=$(kubectl -n block-node get statefulset -o name | head -1)` for the cases below.

**UAT-1 — Fresh install (the reported bug).** Run `block node install --traffic-shaping-enabled …` with
the observer already running. Record pod-ready time, daemon start
(`systemctl show solo-provisioner-daemon -p ActiveEnterTimestamp`) and the first non-empty `publisher={…}`.

*Expected:* membership within **seconds** of pod-ready. *Before the fix:* pod-ready + ~5 min.

**UAT-2 — Daemon restart with no ready pod (fast loop, ~2 min).**

```bash
kubectl -n block-node scale $STS --replicas=0
kubectl -n block-node wait --for=delete pod -l app.kubernetes.io/name=block-node-server --timeout=120s
sudo systemctl restart solo-provisioner-daemon; echo "T0=$(date +%H:%M:%S)"
sudo nft flush set inet weaver-workload-policy bn-publisher
sudo nft flush set inet weaver-workload-policy bn-partner-out
kubectl -n block-node scale $STS --replicas=1
```

*Expected:* `publisher={…}` within seconds of `ready=True`, **not** at `T0 + 5 min`. This is the primary
gate — it forces the daemon to start with no endpoint, which is the race.

**UAT-3 — Control: convergence must stop tracking `poll_interval`.** Repeat UAT-2 with
`components.block_node.statusz.poll_interval` set to `30s`, then `5m`, in
`/opt/solo/weaver/config/daemon.yaml` (restart the daemon — there is no hot-reload).

*Expected:* convergence latency is **the same** in both runs and tracks pod-ready. *Before the fix* it
tracked the interval — that correlation was the diagnostic that proved the bug, so it disappearing is the
proof it is fixed.

**UAT-4 — Pod reschedule changes the endpoint.** With everything converged,
`kubectl -n block-node delete pod -l app.kubernetes.io/name=block-node-server`. The replacement usually
gets a new pod IP.

*Expected:* `journalctl -u solo-provisioner-daemon` shows `TrafficShaperStatuszDiscovered` with the new
`statusz_url`, then `TrafficShaperStatuszEndpointResolved` and a reconcile within seconds — not one
interval later. Membership re-populates. Also confirm the `$VETH` HTB re-attaches (`tc class show dev <veth>`).

**UAT-5 — `base_url` path unregressed.** Set `statusz.base_url` to the pod IP endpoint, restart the daemon,
repeat UAT-2.

*Expected:* unchanged prompt convergence. The override short-circuits `effectiveStatuszURL()` before the
discovered URL is read, so signals are inert on this path.

**UAT-6 — Endpoint loss is logged promptly.** `kubectl -n block-node scale $STS --replicas=0`.

*Expected:* `TrafficShaperStatuszEndpointLost` within seconds, not up to one interval later. No exec fires
(the woken reconcile no-ops on an empty URL) — confirm no `reconcile-shaper` invocation in the journal.

**UAT-7 — No signal storm / no redundant applies.** Leave the node idle and converged for ~15 min.

*Expected:* one `--check` probe per tick, and **no** privileged apply while the roster is unchanged (the
digest gate holds). Only the hourly force-resync should apply. Confirms extra wake-ups are cheap.

### Risks / rollback

- **Signal storm** — a pod flapping ready/not-ready could wake the loop repeatedly. Bounded by the
  capacity-1 buffer (coalescing) and the digest gate (no apply when unchanged), so the worst case is one
  unprivileged `--check` probe per distinct transition. UAT-7 covers this.
- **Send on nil / closed channel** — the channel is never closed and the send is non-blocking; the nil case
  is guarded for direct-construction in tests.
- **Rollback** — revert the commit. No on-disk format, config, or CLI surface changes; behaviour returns to
  waiting for the tick.

Not covered here: `GET /block_node/traffic_shaper/status` reports `{"monitor":{"state":"running"}}` even
while a responsibility is in permanent back-off — observed during this investigation and filed as field
evidence on #750, which already owns per-responsibility health reporting.

### Related Issues

* Closes #1000

### DIFF
--- a/docs/dev/daemon/traffic-shaper-statusz.md
+++ b/docs/dev/daemon/traffic-shaper-statusz.md
@@ -154,11 +154,22 @@ until a ready BN pod is observed the poll loop idles quietly and starts polling
 as soon as one appears.
 
 **Using pod discovery without `base_url`:** after a daemon restart or host
-reboot, the entry reconcile fires immediately but skips if no URL has been
-discovered yet. Convergence then waits until the pod watcher observes a ready
-BN pod, after which the next ticker tick (up to one poll interval) triggers
-the reconcile. For the lowest post-reboot convergence latency, set `base_url`
-to a stable statusz endpoint (e.g. a node-local port-forward).
+reboot, the entry reconcile fires immediately and skips if no URL has been
+discovered yet — the pod watcher runs concurrently, so on a cold start it has
+usually not recorded an endpoint yet. Convergence is **not** deferred to the next
+tick in that case: the watcher signals the poll loop as soon as it records an
+endpoint, and the loop wakes on that signal as well as on the ticker. Both paths
+therefore converge as fast as the BN statusz endpoint responds — bounded by BN
+startup time, not by `poll_interval`.
+
+The same signal fires when a rescheduled pod changes the endpoint, and when the
+owning pod is deleted (the loop wakes, observes an empty URL, and logs the loss
+promptly instead of up to one interval later).
+
+> Before #1000 convergence on this path was bounded by *pod discovery plus up to
+> one poll interval*, because the ticker was the loop's only wake-up source.
+> Setting `base_url` was the documented workaround; it is no longer needed to get
+> prompt convergence.
 
 ### Enablement at install time
 
--- a/docs/dev/traffic-shaper.md
+++ b/docs/dev/traffic-shaper.md
@@ -279,7 +279,7 @@ fill in the parts that are deliberately **not** persisted (see below).
 | Artifact | Persisted at boot? | Rebuilt by |
 |---|---|---|
 | nft tables, chains, rules (both tables) | Yes — replayed from the `.nft` files via `nft -f` | — |
-| nft **set elements** (the CIDR membership of `bn-*` sets) | **No** | daemon statusz poll loop — entry reconcile fires immediately on daemon start; bounded by BN startup time when `base_url` is set, or by pod readiness + up to one poll interval with pod discovery |
+| nft **set elements** (the CIDR membership of `bn-*` sets) | **No** | daemon statusz poll loop — entry reconcile fires immediately on daemon start, and the pod watcher wakes the loop the moment it discovers the endpoint, so convergence is bounded by BN startup time on both the `base_url` and pod-discovery paths |
 | `$EGRESS` HTB hierarchy | Yes — the `solo-provisioner-bandwidth-shaper.sh` script | — |
 | `$VETH` (per-pod) HTB hierarchy | **No** | daemon pod-lifecycle watcher, on the next pod-create event |
 
--- a/internal/daemon/blocknode/pod_watcher.go
+++ b/internal/daemon/blocknode/pod_watcher.go
@@ -222,13 +222,23 @@ func (m *TrafficShaperMonitor) handlePodDelete(ctx context.Context, pod *corev1.
 	// Drop the discovered statusz endpoint only if this is the pod that set it, so
 	// a stale delete for some other pod cannot blank a still-valid endpoint. The
 	// poll loop then idles until another ready BN pod is observed.
-	if m.discoveredStatuszPod == pod.UID {
+	endpointCleared := m.discoveredStatuszPod == pod.UID
+	if endpointCleared {
 		m.discoveredStatuszURL = ""
 		m.discoveredStatuszPod = ""
 	}
 	veth, ok := m.attached[pod.UID]
 	delete(m.attached, pod.UID)
 	m.mu.Unlock()
+
+	// Wake the poll loop so it logs the endpoint loss promptly instead of up to a
+	// full interval later. The woken reconcile is a no-op while the URL is empty.
+	// Signalled after the unlock, and before the early return below so a pod with
+	// no recorded veth still reports the loss.
+	if endpointCleared {
+		m.signalURLChanged()
+	}
+
 	if !ok {
 		return
 	}
@@ -282,6 +292,9 @@ func (m *TrafficShaperMonitor) recordDiscoveredStatusz(pod *corev1.Pod) {
 			Str("pod", pod.Namespace+"/"+pod.Name).
 			Str("statusz_url", url).
 			Msg("discovered BN statusz endpoint from pod")
+		// Wake the poll loop so membership converges now instead of at the next
+		// tick. Sent after the unlock above — never while holding m.mu.
+		m.signalURLChanged()
 	}
 }
 
--- a/internal/daemon/blocknode/pod_watcher_test.go
+++ b/internal/daemon/blocknode/pod_watcher_test.go
@@ -65,10 +65,11 @@ func (f *fakeDelegator) ReconcileShaperCheck(context.Context, string) (string, e
 
 func newTestMonitor(r vethResolver, d *fakeDelegator) *TrafficShaperMonitor {
 	return &TrafficShaperMonitor{
-		resolver:  r,
-		delegator: d,
-		attached:  make(map[types.UID]string),
-		inflight:  make(map[types.UID]bool),
+		resolver:   r,
+		delegator:  d,
+		attached:   make(map[types.UID]string),
+		inflight:   make(map[types.UID]bool),
+		urlChanged: make(chan struct{}, 1),
 	}
 }
 
@@ -303,6 +304,56 @@ func TestHandlePodDelete_ClearsDiscoveredStatuszForOwningPod(t *testing.T) {
 	m.handlePodDelete(context.Background(), pod)
 	require.Empty(t, m.discoveredStatuszURL, "owning pod delete clears the discovered endpoint")
 	require.Equal(t, types.UID(""), m.discoveredStatuszPod)
+}
+
+// signalled reports whether a wake-up is pending, consuming it so successive
+// assertions in one test observe only new signals.
+func signalled(m *TrafficShaperMonitor) bool {
+	select {
+	case <-m.urlChanged:
+		return true
+	default:
+		return false
+	}
+}
+
+// TestRecordDiscoveredStatusz_SignalsOnlyOnChange verifies the poll loop is woken
+// when the endpoint actually changes, and is left alone when a repeat event
+// records the same URL — otherwise every watch event would cost a probe (#1000).
+func TestRecordDiscoveredStatusz_SignalsOnlyOnChange(t *testing.T) {
+	m := newTestMonitor(&fakeResolver{results: []resolveResult{{veth: "lxc1"}}}, &fakeDelegator{})
+	pod := readyPodWithNet("u1", "bn-0", "10.1.2.3",
+		corev1.ContainerPort{Name: bnHealthPortName, ContainerPort: 40983})
+
+	m.recordDiscoveredStatusz(pod)
+	require.True(t, signalled(m), "first discovery wakes the poll loop")
+
+	m.recordDiscoveredStatusz(pod)
+	require.False(t, signalled(m), "an unchanged endpoint must not wake the loop")
+
+	// A rescheduled pod on a new IP is a change and must wake the loop so the
+	// roster is re-read against the new endpoint.
+	moved := readyPodWithNet("u2", "bn-0", "10.1.2.9",
+		corev1.ContainerPort{Name: bnHealthPortName, ContainerPort: 40983})
+	m.recordDiscoveredStatusz(moved)
+	require.True(t, signalled(m), "a changed endpoint wakes the poll loop")
+}
+
+// TestHandlePodDelete_SignalsWhenEndpointCleared verifies losing the endpoint also
+// wakes the loop, so the endpoint-lost transition is logged promptly rather than
+// up to a full poll interval later. An unrelated pod delete must stay silent.
+func TestHandlePodDelete_SignalsWhenEndpointCleared(t *testing.T) {
+	m := newTestMonitor(&fakeResolver{results: []resolveResult{{veth: "lxc1"}}}, &fakeDelegator{})
+	pod := readyPodWithNet("u1", "bn-0", "10.1.2.3",
+		corev1.ContainerPort{Name: bnHealthPortName, ContainerPort: 40983})
+	m.recordDiscoveredStatusz(pod)
+	require.True(t, signalled(m), "drain the discovery signal")
+
+	m.handlePodDelete(context.Background(), readyPod("u2", "bn-1"))
+	require.False(t, signalled(m), "an unrelated pod delete must not wake the loop")
+
+	m.handlePodDelete(context.Background(), pod)
+	require.True(t, signalled(m), "clearing the endpoint wakes the loop")
 }
 
 func TestHandlePodDelete_KeepsDiscoveredStatuszForOtherPod(t *testing.T) {
--- a/internal/daemon/blocknode/traffic_shaper_monitor.go
+++ b/internal/daemon/blocknode/traffic_shaper_monitor.go
@@ -89,6 +89,21 @@ type TrafficShaperMonitor struct {
 	// a delete only clears the endpoint when it is that pod (not some other pod in
 	// a multi-pod window) that went away.
 	discoveredStatuszPod types.UID
+
+	// urlChanged wakes the statusz poll loop when the pod watcher discovers,
+	// changes, or loses the statusz endpoint, so convergence is not deferred to
+	// the next ticker tick. The two responsibilities start concurrently, so the
+	// entry reconcile usually runs before the watcher's initial list has recorded
+	// an endpoint; without this signal the loop would sleep a full poll interval
+	// with a ready pod and a reachable statusz (#1000).
+	//
+	// Capacity 1 with a non-blocking send is load-bearing: it makes the signal
+	// order-independent. A send that lands before the loop reaches its select is
+	// retained in the buffer and returns immediately on arrival, whereas an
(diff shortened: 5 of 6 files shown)


brunodam (approved):
LGTM

